PR #17, opened 2016-01-05
New translation according to v1.0.18

warlof:


src/lang/fr/seat.php
```diff
@@ -397,6 +399,14 @@
     'reinforce_estimate'           => 'Estimation du renforcement',
     'module'                       => 'Module|Modules',
     'content'                      => 'Contenu',
+    'no_known_assets'               => 'Aucuns biens',
```
leonjza: Oops. Indent is out :D

PR #18, opened 2016-01-05
Fix issue eveseat/issues#38

warlof:
This commit fix the wishlist #38 started here https://github.com/eveseat/seat/issues/38

### DIFF
--- a/src/resources/views/includes/header.blade.php
+++ b/src/resources/views/includes/header.blade.php
@@ -20,22 +20,22 @@
 
         <!-- Queue information -->
         <li class="dropdown">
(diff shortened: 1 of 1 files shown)


leonjza:
Dont you think we should refactor this to have something like?

```html
<a href="@if (auth()->user()->has('queue_manager')) {{ route() }} @else #"></a>
```

warlof:
I found this more readable, but if you prefer to do like this, i'll fix it this evening.

leonjza:
Not not very obvious that only the `href` tags are modified like this.

warlof:
done with a ternar =)

leonjza:
Muuuuuch nicer :D

PR #19, opened 2016-01-05
German Language / Typo fix EN

Cynabal:
Suqshed? i hope :D

Typo fix

Adding German

Update seat.php

Update seat.php

Update


Cynabal:
I think i got it :yum:

warlof:
Yup, congrats ^^

PR #22, opened 2016-02-22
Add roles with inverted affiliations

torlivar:
This is a suggestion on how we could make affiliations more flexible. This would allow to make roles with "inverted affiliations" - meaning excludes instead of includes. All affiliations in a role is an exclude if this is checked (which means if you want to combine includes and excludes, you need to make two different roles and assign both to the users).

An accompanying PR for the Service repo is also needed.


leonjza:
Very nice and thank you! =)

I'll take some time soon™ to test this out properly and provide feedback. Are you on Slack btw?

PR #25, opened 2016-06-25
Added total values to the Bounty Prizes and PI monthly ledgers.

jpbrowning:
Generate the total value of monthly Bounty Prize and PI ledger within the view as a panel-footer.

cmusixt

### DIFF
--- a/src/resources/views/corporation/ledger/bountyprizesbymonth.blade.php
+++ b/src/resources/views/corporation/ledger/bountyprizesbymonth.blade.php
@@ -45,6 +45,7 @@
           </thead>
           <tbody>
 
+			      <?php $totalbounty=0; ?>
             @foreach ($bountyprizedates as $bpbm)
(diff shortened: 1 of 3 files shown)


leonjza:
Thanks for the pr!

Although the implementation seems technically correct in calculating the totals, I would prefer us using the `Collection` instance that we get in the view returned from the controller. This will help us not having to use `<?php` tags and reduces the actual LOC needed for this :)

Footers can be added as follows (as an example for `bountyprizesbymonth.blade.php`):
```php
<div class="panel-footer">
   <h3 class="panel-title">Total: {{ number($bountyprizedates->sum('total')) }}</h3>
</div>
```

If you can update the PR then I can merge.

jpbrowning:
Yes, that is much better. Commits updated.

leonjza:
Thank you!

PR #29, opened 2016-07-10
Add corporation titles to character sheet with color parsing

warlof:
Required PR from services repository about character corporation title


warlof:
remove regex and squash both commits

PR #33, opened 2016-08-07
Add collateral amount into contract list

warlof:
Answer to wish eveseat/seat#68


leonjza:
I think we should only show a value if its a courier contract. @warlof can you update the pr for that?

warlof:
OP success, done :)

PR #37, opened 2016-09-09
add alliance icon and name into character list

warlof:
add alliance icon and name into character list according to wish https://github.com/eveseat/seat/issues/134

### DIFF
--- a/src/resources/views/character/list.blade.php
+++ b/src/resources/views/character/list.blade.php
@@ -77,6 +77,7 @@
           <tr>
             <th>{{ trans_choice('web::seat.name', 1) }}</th>
             <th>{{ trans_choice('web::seat.corporation', 1) }}</th>
+            <th>{{ trans('web::seat.alliance') }}</th>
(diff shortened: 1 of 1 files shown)


src/resources/views/character/list.blade.php
```diff
@@ -77,6 +77,7 @@
           <tr>
             <th>{{ trans_choice('web::seat.name', 1) }}</th>
             <th>{{ trans_choice('web::seat.corporation', 1) }}</th>
+            <th>{{ trans('web::seat.alliance') }}</th>
```
warlof: Add alliance column into key table

src/resources/views/character/list.blade.php
```diff
@@ -95,6 +96,12 @@
               {!! img('corporation', $character->corporationID, 64, ['class' => 'img-circle eve-icon small-icon']) !!}
               {{ $character->corporationName }}
             </td>
+            <td>
+              @if($character->allianceID != null)
```
warlof: show the alliance image only if the character is linked to an alliance

PR #38, opened 2016-09-29
Fix permission issues

warlof:
update permission list according to their usage in routes
use ledger permission for corporation ledger routes instead of journal permission

impact : people may have to review a few user/roles permissions.


leonjza:
Thanks! 👍

PR #47, opened 2016-11-13
fix datatable ordering for corporation journal and transaction

warlof:


leonjza:
Fixed globally in https://github.com/eveseat/web/commit/94a6c329bd27be319898d4e96fd919a103723528.

PR #54, opened 2016-12-14
fix issue related to permission

warlof:
fix issue related to permission when an object owner get all permission, including system permission.

### DIFF
--- a/src/Acl/AccessChecker.php
+++ b/src/Acl/AccessChecker.php
@@ -235,10 +235,30 @@ trait AccessChecker
 
             foreach ($key->characters as $character) {
 
-                if ($key->info->type === 'Corporation')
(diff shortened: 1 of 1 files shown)


warlof:
If the content is matching to an owning one (ie : character or corporation like it's the case when you go on your character page), we're set the granted permission to wildcard (*).

It's working well for simple access, but it's not the case for granularity. Yes, when you go on your own stuff, you've access to everything BUT, everything related to this stuff :
mean, if you're on a character stuff, we shouldn't grant corporation permission or anything else

It's explaining why you get access to system permission.

Finally, this issue is not only true for character, but as explained, for corporation aswell.

leonjza:
Nice! Thanks. This should fix eveseat/seat#130. 👍

PR #61, opened 2017-01-21
Fix ACL issue on menu display

warlof:
Provide permission affiliation setting into menu composer in order to fix issue when an user go on character or corporation view.


leonjza:
Awesome thanks! 👍

PR #63, opened 2017-01-28
Fix null exception on page which require permission for unauth user

warlof:
Redirect unauth user to authentication page instead throwing a null exception.


leonjza:
Do you have a way to replicate the bug this fixes?

warlof:
will be fix at routes level

PR #64, opened 2017-02-01
Allow to add custom content from a child template to <head>

BenHUET:
Allow a child template to add content to the `<head>` of the main layout by using the `push` directive. Useful to add  a custom css for example :

(example.blade.php)
```
[...]
@push('head')
  <link rel="stylesheet" href="{{ asset('a/custom/stylesheet.css') }}">
@endpush


leonjza:
I think there are a few people that are going to like this. Thanks for helping make SeAT more awesome! 👍

PR #65, opened 2017-02-05
Improved Tools>People Groups layout

BenHUET:
I found the feature awesome but hard to navigate and I think this rework is a bit clearer than the current one.

Current : 
old

After : 
new

Still no luck with DataTables tho since it does not handle rowspans in the tbody tag.


leonjza:
Oooh *very* nice thanks! 👍 Ill test as soon as possible!
People groups are going to be a *very* important feature when ESI comes into full swing.

PR #70, opened 2017-02-19
Modified cron statements to proper format. Changed from 6 positions to 5

moe-alabel:
Example: from */5 * * * * *  to */5 * * * *


leonjza:
Thanks for helping make SeAT more awesome 👍

PR #71, opened 2017-03-02
Added a "Corporation Characters" Option to Acl

SauerRam:

PR #74, opened 2017-03-06
Sso patch

SauerRam:


leonjza:
What is being patched here? I also think this PR is mixed with eveseat/web#71.

SauerRam:
An existing account triggering the sso activation in user profile will now update the account, not create an new one with the username of the character. So an existing custom Username account can now be sso enabled

warlof:
This fix will allow people with standard account to sign in with an eve account which has a different character name than the account name, right ?

Maybe a better fix could be to allow sso account to have a n:1 relation with classic account instead a 1:1 ?

Because actually, people can only link one eve account to one classic account whereas most of people have at least 2 characters on eve.

Maybe switch the sso credentials to people table and make the check on email address.
It also mean that people who want to sign in without an existing classic account have to provide an email address during authentication process.

leonjza:
Going to hold off on this for now. With ESI coming in to play, the account creation is probably going to be EVE SSO only anyways.

Thanks for you contributions in helping make SeAT more awesome though! 👍

PR #75, opened 2017-04-10
Add item and client filtering to wallet transactions

GunfighterJ:
Previously wallet transaction filtering would only filter by client name. This PR adds filtering for both client name and item name.


warlof:
Remember to add an index on the table on the clientName field ;)

GunfighterJ:
After further testing this doesn't work. It includes results from all characters. I will revise and recreate.

PR #80, opened 2017-06-24
improve tables usability by adding paginate on both header and footer

warlof:
improve tables usability by adding paginate on both header and footer as well as filter feature


leonjza:
🎉

PR #82, opened 2017-08-05
add status filters on industry tables for both character and corporation

warlof:
This improve industry view by adding filter toolbar and replace startDate by endDate

Request eveseat/seat#184
Depends on eveseat/eveapi#21

image

image


warlof:
fix responsive issue

image

PR #84, opened 2017-08-06
allow SeAT owner to overload theme using two custom css file

warlof:
- custom-layout-mini.css is used on login page
- custom-layout.css is used everywhere else

1) create one of the css file or both at the SeAT root directory
2) run `php artisan vendor:publish` in order to publish them

This feature must be used by user and not integrator which can access to header with `head stack`


warlof:
fixed according to discussion

publish step is non longer required

user have to put `custom-layout-mini.css` into `/var/www/seat/public` directory
and/or
user have to put `custom-layout.css` into `/var/www/seat/public` directory

PR #92, opened 2017-10-09
security hotfix which is providing a control on disabled account

warlof:


leonjza:
This confused the activation status based on emails vs account status.

PR #124, opened 2018-04-07
upgrade standing builder to ESI

warlof:
fix eveseat/seat#253


leonjza:
Was this tested? I can't import manually specified characters/corps nor import from other characters.

PR #127, opened 2018-04-10
Align assets representation to ingame experience

herpaderpaldent:
* Show Ship-/Containername + TypeName in parenthesis
* Show only TypeName for mods/ammo etc.


leonjza:
Thanks 👍

PR #134, opened 2018-04-23
fix issue on admin link generation since admin is not a character

warlof:


leonjza:
Thanks, fixed this in 175954d62178ae8956a37a05a53a9ad0fbad7369 rather.

PR #147, opened 2018-05-07
Fix acl groupingissue

warlof:
provide proper migration scripts and fix sidebar issue


leonjza:
We in beta. Re-installs are ok :)

PR #155, opened 2018-05-10
fix doc link to the new amazing documentation ♥

warlof:
address eveseat/seat#294


herpaderpaldent:
This could be impacted with https://github.com/eveseat/docs/pull/32

leonjza:
Parking until we land https://github.com/eveseat/docs/pull/32, just to make sure we don't forget to update the final URL.

PR #187, opened 2018-06-25
Move action buttons to the right

herpaderpaldent:
image


mmolitor87:
While you're at it, switch the edit and delete button positions so accidental presses from impersonate are less likely, especially on mobile. Or move impersonate all the way left - I'm not sure which looks better.

herpaderpaldent:
image

PR #189, opened 2018-06-26
Add overview box with total mined isk

herpaderpaldent:
image


leonjza:
This we are missing a services pr with this for `getTotalCharacterMiningIsk()`.

herpaderpaldent:
BTW: this still leaves open the improvement on the value table ... How i would solve it: create a column with value per row and as long as the time of the row is <23:59 the value should be updated. That should be a pareto efficient way. 
Although it depends when the customer sells his ore and if he get the same value as he had on the day. So any ISK value we show will be under subject of constant change

PR #193, opened 2018-07-01
User list redesign

warlof:
Redesign the user list to something more ergonomic and shining.
- Large button inline for small devices; small button vertical stacked for large devices.
- Hiding character avatar on small devices (except main one)
- Using a list to display character information

ui-redesign-lg
ui-redesign-md
ui-redesign-sm
ui-redesign-xs


warlof:
replaced with eveseat/web#197

PR #194, opened 2018-07-04
Include Main to MemberTracking

herpaderpaldent:
* temp. commented ship-type due to missing location.
* Requires: https://github.com/eveseat/eveapi/pull/107 

https://github.com/herpaderpaldent/eveapi
image


src/resources/views/corporation/tracking.blade.php
```diff
@@ -37,12 +39,22 @@
                 {!! img('character', $character->character_id, 64, ['class' => 'img-circle eve-icon small-icon']) !!}
                 <span rel="id-to-name">{{ $character->character_id }}</span>
               </a>
+            </td>
+            {{--<td>
               @if(!is_null($character->ship_type_id))
                 <i class="pull-right" data-toggle="tooltip"
                    title="" data-original-title="{{ $character->type->typeName }}">
                   {!! img('type', $character->ship_type_id, 64, ['class' => 'img-circle eve-icon small-icon']) !!}
```
leonjza: Please restore this.

src/resources/views/corporation/tracking.blade.php
```diff
@@ -37,12 +39,22 @@
...
                 </i>
               @endif
+
+            </td>--}}
+            <td data-order="{{ optional(optional(optional($character->user)->group)->main_character)->character_id }}>
+              <a href="{{ route('character.view.sheet', ['character_id' => optional(optional(optional($character->user)->group)->main_character)->character_id]) }}">
+                {!! img('character', optional(optional(optional($character->user)->group)->main_character)->character_id, 64, ['class' => 'img-circle eve-icon small-icon']) !!}
```
leonjza: The nested optionals here are :(. We need a better way to do this.

herpaderpaldent: The issue really comes with who is in seat and who isn't. Somehow you need to chain it. I was looking for alternatives, but failed.

src/resources/views/corporation/tracking.blade.php
herpaderpaldent: I came to the conclusion that the ship type is useless without the `last known location`. Therefore i have commented this part to later when we fixxed the location tracking.

image

I have screenshotted if i would add this and find this confusing

src/resources/views/corporation/tracking.blade.php
herpaderpaldent: @leonjza i removed all the optional dependencies and added the ability to asyncronously resolve character_id's for their main_character in one line: 
`<div class="character-id-to-main-character">{{ $character->character_id }}</div>`

see this commit: https://github.com/eveseat/web/pull/194/commits/6d1ac5bdcfa3ad3c0d145133c21af3b147bd161e

herpaderpaldent:
This PR would close: https://github.com/eveseat/seat/issues/357

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -122,4 +123,33 @@ public function resolveIdsToNames(Request $request)
...
+            return (int) $id;
+
+        })->each(function ($chunk) use (&$response) {
+
+            $character = User::find($chunk);
+
+            if (is_null($character)) {
+                $maincharacter = null;
+            } else {
+                $maincharacter = $character->group->main_character;
```
warlof: it can be switch to a ternary `$main_character = is_null($character) ? null : $character->group->main_character;`

src/resources/views/partials/maincharacter.blade.php
```diff
@@ -0,0 +1,4 @@
+<a href="{{ route('character.view.sheet', ['character_id' => optional($maincharacter)->character_id]) }}">
```
warlof: As it will lead to an error, we better should not show the link when we don't have any character_id.
Maybe just display either `unknown` or empty field when we don't have a main.

herpaderpaldent:
Included the feedback from @warlof .

image

Honestly i don't understand the merge conflict?

leonjza:
> Honestly i don't understand the merge conflict?

You have changed the indenting of the conflicting lines it seems.

src/resources/views/corporation/tracking.blade.php
```diff
@@ -13,60 +13,61 @@
...
+                @endif
+              </td>
+              <td>
+                <a href="{{ route('character.view.sheet', ['character_id' => $character->character_id]) }}">
+                  {!! img('character', $character->character_id, 64, ['class' => 'img-circle eve-icon small-icon']) !!}
+                  <span rel="id-to-name">{{ $character->character_id }}</span>
+                </a>
+              </td>
+              <td>
+                <span class="character-id-to-main-character">{{ $character->character_id }}</span>
```
warlof: Actually, this is causing a collapse with the current standard `id-to-name` resolving function. In case the existing resolver is call before you own, the main_character will not be rendered properly.

If the standard is using `[rel="id-to-name"]` to trigger the ajax query, it's using `span:contains(character_id)` as replacement anchors.

In order to fix this and meet a bit with html5 good process, we should apply the following change to the existing flow :
 - moving `<span rel="id-to-name">character_id</span>` to `<span rel="id-to-name" data-character-id="character_id">Unknown</span>`
- updating the replacing anchor from `span:contains(character_id)` to `span[rel=id-to-name][data-character-id=character_id`

If you could apply the change and also include the data attribute to your PR; that will be awesome ♥.

https://github.com/eveseat/web/blob/master/src/resources/views/includes/javascript/id-to-name.blade.php#L36-L38

warlof:
All good for me; thanks for updates :)

PR #206, opened 2018-08-12
Include Main to Coporation Mining Ledger

herpaderpaldent:
This addresses: https://github.com/eveseat/seat/issues/391


src/Http/Controllers/Support/ResolveController.php
```diff
@@ -122,4 +123,28 @@ public function resolveIdsToNames(Request $request)
 
         return response()->json($response);
     }
+
+    public function resolveMainCharacter(Request $request)
```
warlof: Can we move this into a common class as it's duplicated here https://github.com/eveseat/web/pull/194/files#diff-efb6a0fe2c9e632d6125118d1f284731R127 ?

Maybe https://github.com/eveseat/web/blob/master/src/Http/Controllers/Support/ResolveController.php ?

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -122,4 +123,28 @@ public function resolveIdsToNames(Request $request)
...
+        $response = collect();
+        // Grab the ids from the request for processing
+        collect(explode(',', $request->ids))->map(function ($id) {
+
+            // Convert them all to integers
+            return (int) $id;
+        })->each(function ($chunk) use (&$response) {
+
+            $character = User::find($chunk);
+            if (is_null($character)) {
```
warlof: it can be switch to a ternary `$maincharacter = is_null($character) ? null : $character->group->main_character;`

src/resources/views/corporation/mining/ledger.blade.php
```diff
@@ -50,6 +51,9 @@
                                 <span rel="id-to-name">{{ $character_id }}</span>
                             </a>
                         </td>
+                        <td>
+                            <div class="character-id-to-main-character">{{ $character_id }}</div>
```
warlof: Can you apply the same logic from https://github.com/eveseat/web/pull/194 here ?

PR #213, opened 2018-08-31
Mining value

herpaderpaldent:
Adapting views to calculate the correct value.
* changed table header from `amount` to `value`
* detailed ledger is now fully server side rendered
* making sure home-tile is still working
* respecting number representation setting

This is step 3 of 3 for resolving: eveseat/seat#341

image

### DIFF
--- a/src/Http/Controllers/Character/MiningLedgerController.php
+++ b/src/Http/Controllers/Character/MiningLedgerController.php
@@ -53,6 +53,7 @@ class MiningLedgerController extends Controller
 
                 $row->quantity = $row->sum('quantity');
                 $row->volumes = $row->sum('volumes');
+                $row->value = $row->sum('value');
 
                 return $row;
             })->flatten();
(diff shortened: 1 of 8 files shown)


herpaderpaldent:
First of all, this PR is older then your recent introduction of content value instead of ore value.
I am a bit upset that the merge did not happen chronologically and that i am now getting blamed. 

Talking from a financial perspektive amount means how many times something esi ts, whilst we repr

herpaderpaldent:
I am from a market perspective not really in agreement with the latest change with containing minerals. 
As for ore, its comptes able and the market agrees on a value, moon goo is not comptes able and hence its value differ no matter the potenial value when it is refined. There is also a refinement deduction you need to take care of. 
To say the least in eve we got a market where buyers and sellers agree on a value. Artificially changing value seems wrong for seat

warlof:
That's a true point of view. But due to the really small historical data the moongo have on its ORE form, returned price are completely crap and it's quite easy to give them a wrong value.

I don't say it will never change, but the change based on ORE content is actually returning a much more exact amount and doesn't impact the common ORE.

herpaderpaldent:
Please @leonjza let me know your opinion on this.

warlof:
Related to eveseat/services#63 and eveseat/eveapi#118

PR #215, opened 2018-09-09
Fix no_corporation_titles translation

herpaderpaldent:


herpaderpaldent:
Which caption? This one? https://github.com/eveseat/web/blob/master/src/lang/en/seat.php#L509

PR #216, opened 2018-09-12
section for right-sidebar added

cryocaustik:
Added section for right-sidebar as per the AdminLTE examples


warlof:
@leonjza actually it's even allowing further hook than only sending the control sidebar. Assuming it's more flex.

cryocaustik:
I left it as a yield to allow more customization of the sidebar, but I think we could lock it down to 

```php
<aside class="control-sidebar control-sidebar-dark">
    @yield('right-sidebar')
</aside>
<div class="control-sidebar-bg">
</div>
```

would that suffice?

leonjza:
@cryocaustik Lets include a new section for this sidebar, add a template we include that contains the skeleton you suggested and yield from there.

PR #218, opened 2018-09-16
missing file "evegate.png"

ASTairov:
/public/web/img/


ASTairov:
you may need to rename this. it leads into a profile on eve forum

warlof:
What about renaming the file to evelogo.png ?
Of course, it needs to update all view requiring it.

warlof:
@ASTairov could you rename the logo to `evelogo.png` and update the link into the blade ?
https://github.com/eveseat/web/blob/master/src/resources/views/character/includes/summary.blade.php#L112

PR #223, opened 2018-10-05
Smarter ID-Resolver.

herpaderpaldent:
* ID resolver doesn't fall flat anymore with if the request contains illegal ids
* Resolver tries first to resolve corporation and character locally before calling ESI
* refactoring of controller.

### DIFF
--- a/src/Http/Controllers/Support/ResolveController.php
+++ b/src/Http/Controllers/Support/ResolveController.php
@@ -100,7 +100,6 @@ class ResolveController extends Controller
 
             $this->resolveIDsfromESI($chunk, $response, $eseye);
 
-
         });
 
         // Grr. Without this, arbitrary things will get replaced as
@@ -218,7 +217,6 @@ class ResolveController extends Controller
             }
 
         }
-
 
     }
 }
--- a/src/Http/Controllers/Support/ResolveController.php
+++ b/src/Http/Controllers/Support/ResolveController.php
@@ -133,14 +133,14 @@ class ResolveController extends Controller
         // TODO CCP WIP : https://github.com/ccpgames/esi-issues/issues/736
         $names = ChrFaction::whereIn('factionID', $ids->flatten()->toArray())
             ->get()
-            ->map(function ($faction){
+            ->map(function ($faction) {
                 return collect([
                     'id' => $faction->factionID,
-                    'name' => $faction->factionName
+                    'name' => $faction->factionName,
                 ]);
             });
 
-        return $this->cacheIDsAndReturnUnresolvedIDs($names,$ids);
+        return $this->cacheIDsAndReturnUnresolvedIDs($names, $ids);
     }
 
     private function resolveCharacterIDsFromSeat($ids)
@@ -150,14 +150,14 @@ class ResolveController extends Controller
         // no unnecessary api calls the request can be resolved internally.
         $names = CharacterInfo::whereIn('character_id', $ids->flatten()->toArray())
             ->get()
-            ->map(function ($character){
+            ->map(function ($character) {
                 return collect([
                     'id' => $character->character_id,
-                    'name' => $character->name
+                    'name' => $character->name,
                 ]);
             });
 
-        return $this->cacheIDsAndReturnUnresolvedIDs($names,$ids);
+        return $this->cacheIDsAndReturnUnresolvedIDs($names, $ids);
     }
 
     private function resolveCorporationIDsFromSeat($ids)
@@ -167,14 +167,14 @@ class ResolveController extends Controller
         // no unnecessary api calls the request can be resolved internally.
         $names = CorporationInfo::whereIn('corporation_id', $ids->flatten()->toArray())
             ->get()
-            ->map(function ($corporation){
+            ->map(function ($corporation) {
                 return collect([
                     'id' => $corporation->corporation_id,
-                    'name' => $corporation->name
+                    'name' => $corporation->name,
                 ]);
             });
 
-        return $this->cacheIDsAndReturnUnresolvedIDs($names,$ids);
+        return $this->cacheIDsAndReturnUnresolvedIDs($names, $ids);
     }
 
     private function resolveIDsfromESI($ids, $eseye)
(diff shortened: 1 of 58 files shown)


src/Http/Controllers/Support/ResolveController.php
```diff
@@ -78,48 +80,143 @@ public function resolveIdsToNames(Request $request)
 
         })->chunk(1000)->each(function ($chunk) use (&$response, $eseye) {
```
warlof: Sounds a good idea. Can we put but excluded entries before chunking per 1000 ?
So we are certain to send optimized query to ESI.

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -78,48 +80,143 @@ public function resolveIdsToNames(Request $request)
...
-            // universe resolver is not working on factions at this time
-            // retrieve them from SDE and remove them from collection
-            // TODO CCP WIP : https://github.com/ccpgames/esi-issues/issues/736
-            $names = ChrFaction::whereIn('factionID', $chunk->flatten()->toArray())
-                ->get();
+            $chunk = $this->resolveFactionIDs($chunk, $response);
 
-            collect($names)->each(function ($name) use (&$response) {
+            // quick break if no more IDs must be resolve by ESI
+            if ($chunk->count() == 0)
```
warlof: rather than counting (which we don't really care about here), prefer isEmpty instruction : https://laravel.com/docs/5.6/collections#method-isempty

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -78,48 +80,143 @@ public function resolveIdsToNames(Request $request)
...
+                return;
 
-                cache([$this->prefix . $name->factionID => $name->factionName], carbon()->addCentury());
-                $response[$name->factionID] = $name->factionName;
+            $chunk = $this->resolveCharacterIDsFromSeat($chunk, $response);
 
-            });
+            // quick break if no more IDs must be resolve by ESI
+            if ($chunk->count() == 0)
```
warlof: rather than counting (which we don't really care about here), prefer isEmpty instruction : https://laravel.com/docs/5.6/collections#method-isempty

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -78,48 +80,143 @@ public function resolveIdsToNames(Request $request)
...
+                return;
 
-            $chunk = $chunk->filter(function ($id) use ($names) {
-
-                return ! $names->contains('factionID', $id);
-            });
+            $chunk = $this->resolveCorporationIDsFromSeat($chunk, $response);
 
             // quick break if no more IDs must be resolve by ESI
             if ($chunk->count() == 0)
```
warlof: rather than counting (which we don't really care about here), prefer isEmpty instruction : https://laravel.com/docs/5.6/collections#method-isempty

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -78,48 +80,143 @@ public function resolveIdsToNames(Request $request)
...
 
-            // Finally, grab outstanding ids and resolve their names
-            // using Esi.
+            $this->resolveIDsfromESI($chunk, $response, $eseye);
+
+        });
+
+        // Grr. Without this, arbitrary things will get replaced as
+        // #System/Corporation in the UI. Infuriating to say the least.
+        $response->forget([0, 2]);
```
warlof: instead this, can we use the id range table to filter character, corporation, alliance, outposts, constellations, regions and types ? https://gist.github.com/a-tal/5ff5199fdbeb745b77cb633b7f4400bb

src/Http/Controllers/Support/ResolveController.php
herpaderpaldent: The actual `query` to ESI is only send in `resolveIDsfromESI`. I tried to stick to @leonjza 's resolver of fractions which was inside the chunk. 
I can try to alter this but there is not much need to imho

src/Http/Controllers/Support/ResolveController.php
herpaderpaldent: i sticked here to what has been there before, but you are right!

src/Http/Controllers/Support/ResolveController.php
warlof: sure, but with your change, it may send a query with less than 1000 IDs, since few of them may have been resolved with database stuff :)

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -49,103 +67,203 @@ class ResolveController extends Controller
...
+            });
 
-            // We don't have this id in the cache. Return it
-            // so that we can update it later.
-            return true;
+        return $this->cacheIDsAndReturnUnresolvedIDs($names, $ids);
+    }
 
-        })->chunk(1000)->each(function ($chunk) use (&$response, $eseye) {
+    private function resolveCharacterIDsFromSeat($ids)
```
leonjza: Lets name this `resolveInternalCharacterIDs`

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -49,103 +67,203 @@ class ResolveController extends Controller
...
+                ]);
+            });
 
-            collect($names)->each(function ($name) use (&$response) {
+        return $this->cacheIDsAndReturnUnresolvedIDs($names, $ids);
+    }
 
-                cache([$this->prefix . $name->factionID => $name->factionName], carbon()->addCentury());
-                $response[$name->factionID] = $name->factionName;
+    private function resolveCorporationIDsFromSeat($ids)
```
leonjza: Rename to `resolveInternalCorporationIDs`

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -49,103 +67,203 @@ class ResolveController extends Controller
...
+     * @param \Illuminate\Support\Collection $names
+     * @param \Illuminate\Support\Collection $ids
+     *
+     * @return \Illuminate\Support\Collection
+     */
+    private function cacheIDsAndReturnUnresolvedIDs(Collection $names, Collection $ids) : Collection
+    {
+        $names->each(function ($name) {
+
+            cache([$this->prefix . $name['id'] => $name['name']], carbon()->addCentury());
```
leonjza: Why this long and not just forever?

herpaderpaldent: i simply reused the caching period that was used before: https://github.com/eveseat/web/blob/master/src/Http/Controllers/Support/ResolveController.php#L114 

i thought you got your reasons not to cache it for forever

PR #227, opened 2018-11-01
Datatables - Issue 472 - Contact view

herpaderpaldent:
This PR requires https://github.com/eveseat/services/pull/67 

and solves https://github.com/eveseat/seat/issues/472


src/resources/views/character/partials/contact-links.blade.php
```diff
@@ -0,0 +1,38 @@
+@if (!in_array($row->contact_type, ['corporation', 'alliance']))
```
warlof: An interesting approach. Is there a way you make it a bit more generic, so it can be used everywhere we use those links ribbon (ie: in intel, summary sidebar, etc...)

src/Http/Controllers/Character/ContactsController.php
```diff
@@ -37,9 +40,58 @@ class ContactsController extends Controller
     public function getContacts(int $character_id)
     {
 
+        $url = url()->current();
```
warlof: what is the purpose of this ?

src/resources/views/character/contacts.blade.php
```diff
@@ -23,91 +23,54 @@
...
+
+@push('javascript')
+
+  <script type="text/javascript">
+
+    $('#contact-table').DataTable({
+      "processing": true,
+      "serverSide": true,
+      "ajax": {
+        url: "{{$url}}"
```
warlof: simply replace the `$url` call by `url()->current()` ?

src/resources/views/character/partials/contact-links.blade.php
herpaderpaldent: i just updated the PR according to your feedback: 
Now the view lies in partials folder, no longer in /characters/partials: you might want to use `return view('web::partials.links', ['id' => $row->contact_id, 'type' => $row->contact_type]);`

herpaderpaldent:
As soon as this PR is merged i'll continue fixing the reported datatables issues.

herpaderpaldent:
Based on some feedback from my recruiters i extended the contacts datatables:
Now you can review All contacts of someones group. Additionally to this contacts in the very same group are shown with a blue background:
image

src/resources/views/character/contacts.blade.php
herpaderpaldent: this was very helpful feedback

src/Http/Controllers/Character/ContactsController.php
```diff
@@ -22,24 +22,119 @@
...
+                return $user_group->intersect(collect($row->contact_id))->isNotEmpty();
+            })
+            ->filterColumn('name', function ($query, $keyword) {
+                $resolved_ids = ResolvedIds::where('name', 'like', '%' . $keyword . '%')->get()->map(function ($resolved_id) { return $resolved_id->id; });
+                $character_info_ids = CharacterInfo::where('name', 'like', '%' . $keyword . '%')->get()->map(function ($character_info) { return $character_info->character_id; });
+                $corporation_info_ids = CorporationInfo::where('name', 'like', '%' . $keyword . '%')->get()->map(function ($corporation_info) { return $corporation_info->corproation_id; });
+
+                $query->whereIn('contact_id', array_merge($resolved_ids->toArray(), $character_info_ids->toArray(), $corporation_info_ids->toArray()));
+            })
+            ->filterColumn('label_ids', function ($query, $keyword) {
```
warlof: You can replace this by the one bellow which will return much proper results and cover all labels

```
->filterColumn('label_ids', function ($query, $keyword) use ($character_ids) {

	$labels = CharacterContactLabel::where('label_name', 'like', '%' . $keyword . '%')
		->whereIn('character_id', $character_ids)
		->get();

	foreach ($labels as $label)
		$query->whereRaw(DB::raw('JSON_CONTAINS(label_ids, ' . $label->label_id . ')'));
})
```

warlof:
Also, if you can add 5 checkbox on the datatable and based search on them for standing columns ?
-10 -5 0 5 10
Having all checked by default.

This will allow user to filter on specific standings only without beeing in conflict with character/corporation/alliance or whatever ID

herpaderpaldent:
Ok this PR should be ready for review:
image

src/WebServiceProvider.php
```diff
@@ -118,14 +120,18 @@ public function add_publications()
...
 
             // Font Awesome Pulled from packagist
             base_path('vendor/components/font-awesome/css/font-awesome.min.css') => public_path('web/css/font-awesome.min.css'),
             base_path('vendor/components/font-awesome/fonts')                    => public_path('web/fonts'),
         ]);
     }
 
+    public function add_migrations()
+    {
+        $this->loadMigrationsFrom(__DIR__ . '/database/migrations/');
```
leonjza: I think this is unrelated to this PR.

PR #229, opened 2018-11-07
Asyncronous `/configuration/users`

herpaderpaldent:
This solves: https://github.com/eveseat/seat/issues/449 

needs: https://github.com/eveseat/services/pull/69 

image


src/Http/Controllers/Configuration/UserController.php
```diff
@@ -42,9 +46,74 @@ class UserController extends Controller
...
-        return view('web::configuration.users.list', compact('groups'));
+            $groups = $this->getAllFullUsers();
+
+            return DataTables::of($groups)
+                ->editColumn('refresh_token', function ($row) {
+                    return view('web::configuration.users.partials.refresh-token', compact('row'));
+                })
+                ->editColumn('name', function ($row) {
+
+                    if($row->id === 1)
```
warlof: we rather should exclude the admin from the users list since it's a built-in user.

src/Models/Group.php
```diff
@@ -88,6 +88,8 @@ class Group extends Model
      */
     protected $fillable = ['main_character_id'];
 
+    protected $appends = ['main_character'];
```
warlof: Can you add the missing documentation block here ?

src/Models/User.php
```diff
@@ -144,6 +144,8 @@ class User extends Model implements AuthenticatableContract, CanResetPasswordCon
      */
     protected $hidden = ['password', 'remember_token'];
 
+    protected $appends = ['email'];
```
warlof: Can you add the missing documentation block here ?

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -42,9 +46,74 @@ class UserController extends Controller
...
+                    if(empty($row->email))
+                        return trans('web::seat.no_email');
+
+                    return $row->email;
+                })
+                ->addColumn('action_buttons', function ($row) {
+                    return view('web::configuration.users.partials.action-buttons', compact('row'));
+                })
+                ->addColumn('roles', function (User $user) {
+                    return $user->group->roles->map(function ($role) {
```
warlof: Is there a way you return the caption `No role defined` as you did for email ?

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -42,9 +46,74 @@ class UserController extends Controller
     public function getAll()
     {
 
-        $groups = $this->getAllGroups();
+        if(\request()->ajax()){
```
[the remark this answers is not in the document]
herpaderpaldent: done

src/Http/Controllers/Configuration/UserController.php
herpaderpaldent: done in service package PR

src/Http/Controllers/Configuration/UserController.php
herpaderpaldent: done

src/Models/Group.php
herpaderpaldent: done

src/Models/User.php
herpaderpaldent: done

src/resources/views/configuration/users/list.blade.php
```diff
@@ -8,124 +8,83 @@
...
+        },
+        dataSrc: 'main_character'
+      },
+      drawCallback : function () {
+        $("img").unveil(100);
+        ids_to_names();
+      },
+    });
+
+    function wordwrap( str, width, brk, cut ) {
```
leonjza: I noticed the use of this function again, another candidate for `seat.js` I think. Could you maybe update the other PR's that have this too? This way we have a single origin for the code, less copy paste and have more maintainable code :D

PR #231, opened 2018-11-08
Add ability to remove sold / bio massed characters from user_group

herpaderpaldent:
* will create an intel note entry
* intel notes are now showed in `/configuration/users/edit/`

This completes the following issue: https://github.com/eveseat/seat/issues/449 

image


herpaderpaldent:
i shall add that this PR is not aimed to over engineer the whole thing. I wasn't able to update the `main_character_id` of a group if you deactivate the main_character. This would be a super edge case and could be solved afterwards with impersonating. 
It's a very functional solution to aim with sold and bio massed characters. Moving them out of a user group to prevent `MissingRefreshToken` Exceptions (relevant for SeAT-Groups and Discord-Connector)

herpaderpaldent:
Hey @warlof 

image

i included our yesterdays discussions, fixed an persisting error from 2.x migration (set-email) and added the ability to activate an user again. I did add the ability to lock seat from registration too:

src/resources/views/auth/login.blade.php
```diff
@@ -25,5 +26,12 @@
 
   </div>
   <!-- /.login-box-body -->
+  @else
```
warlof: How should we login if we already have an account ?

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -75,17 +79,14 @@ public function editUser($user_id)
...
     {
 
         $user = $this->getUser($request->input('user_id'));
 
-        $user->fill([
-            'email' => $request->input('email'),
-        ]);
-
-        $user->save();
+        Profile::set('email_address', $request->input('email'), $user->group->id);
```
warlof: Please use the helper instead making direct call to this class : `setting(['email_addres', $request->input('email'), $user->group->id]);`

https://github.com/eveseat/services/blob/master/src/Helpers/helpers.php#L267

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -116,13 +117,44 @@ public function postReassignuser(ReassignUser $request)
...
+                    'group_id' => Group::create()->id,
+                ]);
+
+                $user->save();
+
+            } else {
+                $note = collect(sprintf('This character has been deactivated by %s.',
+                    auth()->user()->name));
+            }
+        } else {
```
warlof: This must be the default case. There is no need for an `else` here

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -116,13 +117,44 @@ public function postReassignuser(ReassignUser $request)
...
+                    auth()->user()->name, $current_group->users->map(function ($user) { return $user->name; })->implode(', ')));
+
+                $user->fill([
+                    'group_id' => Group::create()->id,
+                ]);
+
+                $user->save();
+
+            } else {
+                $note = collect(sprintf('This character has been deactivated by %s.',
```
warlof: This must be the default case. There is no need for an `else` here.

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -116,13 +117,44 @@ public function postReassignuser(ReassignUser $request)
...
+        if($user->active){
+
+            if ($current_group->users->count() > 1) {
+
+                $note = collect(sprintf('This character has been deactivated by %s and it formerly belonged to the following user group: %s.',
+                    auth()->user()->name, $current_group->users->map(function ($user) { return $user->name; })->implode(', ')));
+
+                $user->fill([
+                    'group_id' => Group::create()->id,
+                ]);
```
warlof: you can call `save()` directly after the filler.

src/Http/Controllers/Auth/SsoController.php
```diff
@@ -237,4 +229,18 @@ private function updateMainCharacterId(User $user)
...
+
+    /**
+     * If a deactivated user provides a new RefreshToken
+     * the user is reactivated and a note is created.
+     *
+     * @param \Seat\Web\Models\User $user
+     */
+    private function activateUser(User $user)
+    {
+        $user->active = true;
```
warlof: Is there a missing `save` statement ?
Otherwise, that mean you can create a note but keep the user disabled.

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -116,13 +117,44 @@ public function postReassignuser(ReassignUser $request)
...
      * @return mixed
      */
-    public function editUserAccountStatus($user_id)
+    public function editUserAccountStatus(NewIntelNote $request, int $user_id)
     {
 
+        $user = $this->getUser($user_id);
+        $current_group = $user->group;
+
+        if($user->active){
```
warlof: missing spaces after `if` and before the backet.

src/Http/Controllers/Auth/LoginController.php
```diff
@@ -79,6 +81,8 @@ public function showLoginForm()
         if (strlen(env('EVE_CLIENT_SECRET')) < 5 || strlen(env('EVE_CLIENT_ID')) < 5)
             session()->flash('warning', trans('web::seat.sso_config_warning'));
 
-        return view('web::auth.login');
+        $allow_registration = Seat::get('registration') === 'no' ? false : true;
```
warlof: It can be shortened by `$allow_registration = setting('registration', true) !== 'no'`
Furthermore, use helper instead making direct call to this class.

https://github.com/eveseat/services/blob/master/src/Helpers/helpers.php#L267

src/Http/Controllers/Auth/SsoController.php
herpaderpaldent: No this works just fine 
image

src/resources/views/auth/login.blade.php
herpaderpaldent: https://github.com/herpaderpaldent/web/blob/a34205db19eaf93f615a322666d648b70d6e860d/src/Http/Controllers/Auth/SsoController.php#L154-L166

This allows to login into existing accounts. However the error is not presented to the user. I don't know why exactly but i suspect it has to do with the `send()` method. 
However this is needed as if i'd be using return the controller claims that it expects a return statement of the type User. 

Here i am very open for better solutions.

herpaderpaldent: There is another option:
If the registration is disbled show such an information on the auth/login page everytime someone visiting stating f.e. "Registration of new characters has been disabled, however existing users will be able to login just as they are used to".

warlof:
This PR still need a few changes.
 - As a standard user, I'm not able to deactivate/remove the character. It should be the case, at least with a permission or a global setting.
 - A disabled user is still able to navigate into SeAT if he's or not already logged-in => disabled account don't have any effect
 - When a disabled user is also a group main, the settings is not updated. At least, we should keep a remaining random character from the group. The best would be to ask the user which remaining character should become a main.
 - The note table from users edit card need an escaped column for Note field - otherwise, plain html will be displayed instead being rendered.
 - When the registration is disabled and I'm attempting to sign-in with a not registered user, I'm redirected to the login page without any information (what append ?)

src/resources/views/auth/login.blade.php
warlof: Your notification is not displayed because you're not returning it but only exiting the function.
This have to be done in the called method : https://github.com/eveseat/web/blob/ab10fb4c0c8e83c62003ff4d02fab678244f0ec9/src/Http/Controllers/Auth/SsoController.php#L63

You can maybe handle this by checking if both settings is false and `$user` is null

herpaderpaldent:
So this PR had caused some mix feelings for me. We had a quite heated discussion on slack although i have the feeling that generally the cause of this is a misconception of the introduced feature. 

The feature does:

> Add ability to remove sold / bio massed characters from user_group

it does not aim to ban a user from using seat. Therefore:

> A disabled user is still able to navigate into SeAT if he's or not already logged-in => disabled account don't have any effect

This is correct and aimed behavior. However only a sold character is able to do so, as a biomassed character won't be able to receive a `token` via SSO. Disabling a user helps f.e. marking user in the user-list as such a special case: 

https://github.com/eveseat/web/blob/3d862ff4e0860ae97cbb4bfa480d9f4135f73145/src/resources/views/configuration/users/list.blade.php#L74-L77 

> When a disabled user is also a group main, the settings is not updated. At least, we should keep a remaining random character from the group. The best would be to ask the user which remaining character should become a main. 

i have mentioned this in the comment on this PR as i was struggeling with it. I stand that selling/biomassing a character is a super edge chase and could be solved by the admin disabeling it easily.
image

> When the registration is disabled and I'm attempting to sign-in with a not registered user, I'm redirected to the login page without any information (what append ?)

This is correct, i forgot to add this to the view: https://github.com/eveseat/web/pull/231/commits/33bf6af9df50262660a8fbe4d036e45341ee8942
image

> an user come to me and say "this toon has been biomassed"
as a super, I go in users list, disabling the user and saying what the owner told me "sold toon"
the user is now "something"
tomorrow, the owner is sign in with this user
the user is no longer "something"

> what's the purpose ?

As mentioned the purpose of this PR is:

> Add ability to remove sold / bio massed characters from user_group

additionally i stated in the PR's comment:

> It's a very functional solution to aim with sold and bio massed characters. Moving them out of a user group to prevent MissingRefreshToken Exceptions (relevant for SeAT-Groups and Discord-Connector)

As for administration: 

Regarding the biomassed character: CCP is writing an email to the whole corporation if someone biomasses the character. As superuser i’d ask the user to provide me some prove before i’d deactivate the user and add that proof to the intel note.

Regarding the selling use-case my recruiter would do exactly the same: some screenshot or the evemail proofing that it has been sold and to whom it was sold. I’d write that to the intel note.

---

User-Acceptance:

This feature is based from various feedback from my recruiters. We have discussed this in both support and dev channel and my proposal was generally receipted wildly positively. 

f.e. 
image 

---

The amount of discussion and misconception (f.e. thinking that this is a banning feature or that a user could administrate (selling/biomassing) himself) of it has taken a lot of energy of me and i hope this comment finally resolves this topic.

src/Http/Controllers/Auth/LoginController.php
```diff
@@ -23,6 +23,7 @@
 namespace Seat\Web\Http\Controllers\Auth;
 
 use Illuminate\Foundation\Auth\AuthenticatesUsers;
+use Seat\Services\Settings\Seat;
```
leonjza: Possibly unused class to remove. (using helper now)

src/Http/Controllers/Auth/SsoController.php
```diff
@@ -145,6 +151,20 @@ private function findOrCreateUser(SocialiteUser $eve_user): User
             return $existing;
         }
 
+        // Detect if registration is disabled
+        // Abort user creation.
+        if (setting('registration', true) === 'no') {
+
+            // Log the account creation attempt
+            event('security.log', [
+                'Attempt of account creation for ' . $eve_user->name, ' blocked, as registration is turned off.',
```
leonjza: Reword to:

```
Account creation for {user} failed. Registration has been disabled.
```

This should probably also be a translation candidate.

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -181,13 +179,43 @@ public function postReassignuser(ReassignUser $request)
...
-    public function editUserAccountStatus($user_id)
+    public function editUserAccountStatus(NewIntelNote $request, int $user_id)
     {
 
+        $user = $this->getUser($user_id);
+        $current_group = $user->group;
+
+        // Set default note
+        $note = collect(sprintf('This character has been reactivated by %s.',
+            auth()->user()->name));
```
leonjza: I am not 100% on board with this. Adding notes like this will reveal to a standard user who manipulated their account (as well as tell them who else is on the instance). I would prefer to not leak such information to users other than administrators. Instead, the security log should be used for account manipulation events (always), and the character note may be added as convenience without the user who modified it.

src/Http/Controllers/Configuration/UserController.php
```diff
@@ -181,13 +179,43 @@ public function postReassignuser(ReassignUser $request)
...
+
+        if ($user->active){
+
+            // Initial setting of deactivation note
+            $note = collect(sprintf('This character has been deactivated by %s.', auth()->user()->name));
+
+            if ($current_group->users->count() > 1) {
+
+                // If deactivated user is part of a larger user group, use another note
+                $note = collect(sprintf('This character has been deactivated by %s and it formerly belonged to the following user group: %s.',
```
leonjza: Similarly here. Recording the group information will give a standard user more information that I believe they should know. Use the security log feature we already have for that.

src/lang/en/seat.php
```diff
@@ -122,6 +122,7 @@
     'sign_in'               => 'Sign In',
     'email_verify'          => 'Email Verification',
     'email_verify_welcome'  => 'Please enter a valid email address for your account.',
+    'no_register'           => 'Registration has been disabled. Please contact your administrator if you find this erroneous.',
```
leonjza: Reword to:

```
Registration has been disabled. Please contact your administrator.
```

src/resources/views/configuration/users/edit.blade.php
```diff
@@ -29,25 +31,124 @@
...
-               class="btn btn-{{ $user->active ? 'warning' : 'success' }} pull-left">
+            <a type="button" class="btn btn-{{ $user->active ? 'warning' : 'success' }} pull-left" data-toggle="modal" data-target="{{ $user->active ? '#deactivateModal' : '#activateModal' }}">
+
               @if($user->active)
                 {{ trans('web::seat.deactivate_user') }}
               @else
                 {{ trans('web::seat.activate_user') }}
               @endif
             </a>
+
```
leonjza: These new sections need translations please.

src/Http/Controllers/Auth/SsoController.php
herpaderpaldent: maybe in a later PR as it would require to exchange the variable dynamically

src/resources/views/configuration/users/edit.blade.php
herpaderpaldent: maybe we can do this in another PR or you add to this pull request, as i know that i won't match your wording?

src/Http/Controllers/Configuration/UserController.php
herpaderpaldent: i moved the notes to the sec.log and removed the notes.

src/Http/Controllers/Configuration/UserController.php
herpaderpaldent: i moved the notes to the sec.log and removed the notes.

herpaderpaldent:
This pull request is open from the 8th of november. Your PR-Setup sucks @warlof and @leonjza. If one of you chose to be away from the project the other should be able to review and merge changes.

Your behavior in this pull request is a **complete refusal to do your job**. How can both of you have not found time to approve/merge this change, i can not comprehend. 

Will not further work on this. Feel free to reopen it when you change your setup

PR #235, opened 2018-11-19
use laravels `loadMigrationsFrom()` method instead of publishing ass…

herpaderpaldent:
use laravels `loadMigrationsFrom()` method instead of publishing assets

This is based on a review feedback from @warlof:
https://github.com/eveseat/web/pull/233/files#r234420908


leonjza:
Ah nice!

PR #236, opened 2018-11-21
Resolve ids universe names

herpaderpaldent:
Use existing UniverseNames instead of creating ResolvedIds Model.

This would have impact on the following two PRs:
- ~~https://github.com/eveseat/web/pull/233~~ (removed dependency)
- ~~https://github.com/eveseat/web/pull/227~~ (removed dependency)

f.e. 
https://github.com/eveseat/web/pull/233/files#diff-8b412ff1d63c9ac6479a3bde722c4da5 

would become obsolete and name resolving could be done via already existing 'universe_names' table

### DIFF
--- a/src/Http/Controllers/Support/ResolveController.php
+++ b/src/Http/Controllers/Support/ResolveController.php
@@ -27,6 +27,7 @@ use Illuminate\Support\Collection;
 use Seat\Eveapi\Models\Character\CharacterInfo;
 use Seat\Eveapi\Models\Corporation\CorporationInfo;
 use Seat\Eveapi\Models\Sde\ChrFaction;
+use Seat\Eveapi\Models\Universe\UniverseName;
 use Seat\Web\Http\Controllers\Controller;
 use Seat\Web\Models\User;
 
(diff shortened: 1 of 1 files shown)


src/Http/Controllers/Support/ResolveController.php
```diff
@@ -234,6 +266,14 @@ private function cacheIDsAndReturnUnresolvedIDs(Collection $names, Collection $i
             cache([$this->prefix . $name['id'] => $name['name']], carbon()->addCentury());
             $this->response[$name['id']] = $name['name'];
 
+            // Faction is not allowed in UniverseNames table
+            if($name['category'] !== 'faction') {
+                UniverseName::firstOrCreate([
+                    'entity_id' => $name['id'],
```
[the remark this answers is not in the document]
herpaderpaldent: i don't know why i did this wrong. It's not the first time i use this method, but the first time i do it so wrongly. 
Thank you for finding it.

herpaderpaldent:
Ok @warlof PR should now be good to go

src/Http/Controllers/Support/ResolveController.php
```diff
@@ -140,6 +146,24 @@ private function resolveFactionIDs($ids)
...
                     'id' => $faction->factionID,
                     'name' => $faction->factionName,
+                    'category' => 'faction',
+                ]);
+            });
+
+        return $this->cacheIDsAndReturnUnresolvedIDs($names, $ids);
+    }
+
+    private function resolveInternalUniverseIDs($ids)
```
leonjza: Could we get some comments on these functions?

herpaderpaldent: There were already comments on what the function does inside the function. However i have added some php_doc on these functions. I hope that is what you wanted?
https://github.com/eveseat/web/pull/236/commits/b4fc41805b9723d82ce1fa9da4dd3bd6cae2256e

PR #240, opened 2018-12-02
Update market tab and refactor partials

herpaderpaldent:
requires: https://github.com/eveseat/services/pull/74

Update market tab:

image

solves:  https://github.com/eveseat/seat/issues/458


src/Http/Controllers/Character/MarketController.php
```diff
@@ -57,18 +58,37 @@ public function getMarket(int $character_id)
     public function getMarketData(int $character_id)
     {
 
-        $orders = $this->getCharacterMarketOrders($character_id, false);
+        if (! request()->has('all_linked_characters'))
+            return abort(500);
```
warlof: shouldn't rather be a 400 ? @leonjza

src/Http/Controllers/Character/MarketController.php
```diff
@@ -57,18 +58,37 @@ public function getMarket(int $character_id)
...
+
+        $orders = $this->getCharacterMarketOrders($character_ids);
 
         return DataTables::of($orders)
             ->addColumn('bs', function ($row) {
 
-                return view('web::partials.marketbuysell', compact('row'))
-                    ->render();
+                if ($row->is_buy_order)
+                    return '<span class="text-red">Buy</span>';
```
warlof: Can you restore the partial blade ? - so we will keep the controller clean of HTML.

src/Http/Controllers/Character/MarketController.php
```diff
@@ -57,18 +58,37 @@ public function getMarket(int $character_id)
...
+
+                return '<span class="text-green">Sell</span>';
             })
             ->addColumn('vol', function ($row) {
 
-                return view('web::partials.marketvolume', compact('row'))
-                    ->render();
+                if ($row->is_buy_order)
+                    return number($row->volume_total, 0);
```
warlof: I may agree with the number thing, but can you keep it into the blade ?
Also, with the asynchronous working, shouldn't the number feature be moved into a datatable render ?

src/Http/Controllers/Character/MarketController.php
herpaderpaldent: Qu1ckkk said here to use 500: https://eve-seat.slack.com/archives/C0H3VGH4H/p1543645324006200

src/Http/Controllers/Character/MarketController.php
herpaderpaldent: About the numbers: https://github.com/eveseat/web/blame/a1274fa520c2925b3fb5396fa2864d8d0d8fa1f9/src/Http/Controllers/Character/MarketController.php#L75 
That's the way it has been used for quite some time and the way the helper function works, this seems to be right. Doing this asynchronously now, does not have anything to do how the number format is rendered. Either you render it in the controller or by the client. However if you'd let the client render it. We might need to move the whole helper to a js version. This is comparable to the human_readable js helper: 
https://github.com/eveseat/web/blob/master/src/resources/assets/js/seat.js#L67-L75

i have the feeling, editing the column in the asynchronous controller is the right place to do so. And i would suggest keep this as is. 

The partial blade for `<span class="text-red">Buy</span>` i have restored, this i understand

src/Http/Controllers/Character/MarketController.php
herpaderpaldent: yes done: https://github.com/eveseat/web/pull/240/commits/6a011686bda9b58903f6171a18b4713aad84edc6

PR #243, opened 2018-12-06
Mailtimeline Improvement: only link existing characters and corporations

herpaderpaldent:
Only existing characters and corporations are going to be linked. Additionally the id-resolution will be skipped if the ID is already cached

Solves https://github.com/eveseat/seat/issues/428


src/resources/views/partials/character.blade.php
```diff
@@ -1,17 +1,17 @@
 @if (request('all_linked_characters') === "true")
-  {!! img('character', $character_id, 32, ['class' => 'img-circle eve-icon small-icon'],false) !!}
+  {!! img('character', $character_id, 32, ['class' => 'img-circle eve-icon small-icon'], false) !!}
 @endif
 
 @if (isset($character->name) )
```
warlof: should be `@if (isset($character->name))`

herpaderpaldent: yes and mergeconflict resolved

PR #251, opened 2018-12-15
fix(users): Reduce user coupling and avoid null exception

warlof:
When an user is attached to a non existing group, the user list may
crash due to some relation which are returning a null value.

This is reducing coupling for an User model by using directly the group_id from
the user entry and excluding things attached to group when none are
returned (mails, role, etc...).

Closes eveseat/seat#491


warlof:
> 
> 
> I'd propose to fix this in the service package:
> Add: `->has('group')` to eloquent query builder

Rather than hiding shit, it's better to fixing the bugged user account.

PR #254, opened 2018-12-20
Include MailingList name, include source and refactor alliance view

herpaderpaldent:
image

- The source of the mail is declared
- refactor to_alliance part to use alliance partial
- introduce to_mailinglist part.

**Depends**: https://github.com/eveseat/eveapi/pull/148


src/resources/views/character/mail-timeline.blade.php
```diff
@@ -5,6 +5,7 @@
 
 @inject('CharacterInfo', 'Seat\Eveapi\Models\Character\CharacterInfo')
 @inject('CorporationInfo', 'Seat\Eveapi\Models\Corporation\CorporationInfo')
+@inject('MailMailingList', 'Seat\Eveapi\Models\Mail\MailMailingList')
```
warlof: Will make some tries tomorrow, but I don't like those Inject :p
We should avoid them as much as possible.

herpaderpaldent: i know how you feel ... the whole blade does live up to my expectations. However because there were quick fixes to make it usable, i went for those low hanging fruits, until we manage to refactor the whole mail-timeline.

src/resources/views/character/mail-timeline.blade.php
```diff
@@ -77,7 +77,26 @@
...
+
+              {{ isset($MailMailingList::where('mailing_list_id', $recipient->recipient_id)->first()->name) ? $MailMailingList::where('mailing_list_id', $recipient->recipient_id)->first()->name : trans('web::seat.unknown') }}
+
+            @endforeach
+
+          </li>
+
+        @endif
+
+        <b>{{ trans('web::seat.source') }}: </b>
```
warlof: you forget to push the language file.

herpaderpaldent: i didn't need to as that is already in the language file: https://github.com/eveseat/web/blob/master/src/lang/en/seat.php#L359

src/resources/views/character/mail-timeline.blade.php
```diff
@@ -77,7 +76,30 @@
...
+              @endif
+
+            @endforeach
+
+          </li>
+
+        @endif
+
+        <b>{{ trans('web::seat.source') }}: </b>
+        @include('web::partials.character', ['character' => $CharacterInfo::find($message->character_id) ?: $message->character_id, 'character_id' => $message->character_id])
```
leonjza: Sorry, but model look ups in views are not a type of complexity I think we should be adding.

src/resources/views/character/mail-timeline.blade.php
```diff
@@ -77,7 +76,30 @@
...
+                {{ trans('web::seat.unknown') }}
+              @endif
+
+            @endforeach
+
+          </li>
+
+        @endif
+
+        <b>{{ trans('web::seat.source') }}: </b>
```
leonjza: Is this source the character that sent to the mailing list? Why don't we just update the From with that?

herpaderpaldent: it's the data source is not equal to the sender. As the source might be either sender or receiver. The main reason to include this information are that HR scans the Mail-timeline and might be interested which of all the characters they have access to had provided the mail-information as they might be in corporation which acts as receiver.

src/resources/views/character/mail-timeline.blade.php
herpaderpaldent: overall i agree, however the whole view and controller for mail-timeline needs to have a proper refactoring. This model look ups are already across this view and would not add to its complexity. It is however a quick application to add value to the view without putting in to much effort. 

Having said that. I am aware that this view needs a redo as the way it is build up is way to complex.

src/resources/views/character/mail-timeline.blade.php
herpaderpaldent: in the PR's picture you can see how 'from' 'to' and 'source' can be totally different.

leonjza: Is the source the account that actually has this email?

herpaderpaldent: yes 
image

leonjza: We need a different name than source. I also want to suggest; as this is a meta-like piece of information, maybe it should move to the right under the new title and not be part of the headers section but rather timestamps.

Lastly, did you check that we won't have a case where two people on the same SeAT instance, with the same email wont disclose the presence of these accounts on the instance with this feature?

herpaderpaldent: I moved the source to the right below the timestamp:
image
 
Until you have better suggestion i stay with source as name, as it appears to be a good terminology in regards to mail transactions. From or origin may be misleading.

Regarding the exposing of characters and corporations on same seat instance: this is already a fact in not only this blade but any Datatable be it wallet transactions, intel page, contacts etc. In this blade `from` and `to` already expose the existance. IMHO this should not block this PR.

Furthermore as source is the character_id attribute from the table: The source of an email, to an auth user must always be affiliated with the auth()->user because this is quaranteed in the service:
https://github.com/eveseat/services/blob/ffba218379d7aff0e8fec52b8c23c0a4a03fb3cb/src/Repositories/Character/Mail.php#L141

herpaderpaldent:
Heyho

I removed the class injection as requested by @leonjza. Now a new dependency is needed: https://github.com/eveseat/eveapi/pull/148

herpaderpaldent:
Could this be merged, my recruitment team is asking for it.

herpaderpaldent:
I am closing my open PR's and won't be working on the branch anymore. 

Your setup you have established which requires everything to go over your @leonjza 's desk is not working for me anymore. It should be good enough if one reviewers approves it.

PR #256, opened 2018-12-22
Name title view

warlof:
Upgrade both character and corporation layouts in order to introduce the entity name into the page `<title>` tag.

Introduce a new layout parameter called `breadcrumb` which will be used in order to pass the blade title to the page title. The entity name is sent via a View Composer.

Closes eveseat/seat#500

### DIFF
--- a/src/Http/Composers/CharacterLayout.php
+++ b/src/Http/Composers/CharacterLayout.php
@@ -1,0 +1,56 @@
+<?php
+/*
+This file is part of SeAT
+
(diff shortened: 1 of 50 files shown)


leonjza:
Any reason you chose `breadcrumb` over `title` here?

warlof:
`title` was already in use while I started to work on this. I then moved a something more abstract.
I can switch from `breadcrumb` to `title` if you prefer.

leonjza:
After the discussion on Slack, breadcrumb is fine 👍

PR #257, opened 2018-12-23
feat(roles): Allow roles name to be changed

warlof:
Closes eveseat/seat#375


src/Http/Routes/Configuration/Access.php
```diff
@@ -40,6 +40,11 @@
     'uses' => 'AccessController@editRole',
 ]);
 
+Route::post('/roles/edit/{id}', [
+    'as'   => 'configuration.access.roles.rename',
+    'uses' => 'AccessController@renameRole',
+])->where('id', '[0-9]+');
```
herpaderpaldent: i have the feeling that we have avoided verification inside the routes file. Personally i would have created a verification for the request.

src/lang/en/seat.php
```diff
@@ -23,6 +23,7 @@
 return [
 
     // General
+    'global'                => 'Global',
```
herpaderpaldent: maybe role would be more appropriate?

src/Http/Routes/Configuration/Access.php
warlof: Indead, the bouncer is set into the parent group https://github.com/warlof/web/blob/b99ee2e0ceb2d04d5a4b595b0ac2225d29b8b979/src/Http/routes.php#L126

herpaderpaldent: i didn't meant bouncer, i meant verification: https://laravel.com/docs/5.5/validation#creating-form-requests

leonjza: This is a simple constraint for a URL parameter that is fine (and quite nice actually).

src/Http/Controllers/Configuration/AccessController.php
```diff
@@ -108,6 +108,20 @@ public function editRole($role_id)
             ));
     }
 
+    public function renameRole($role_id)
```
leonjza: Function comments please.

src/Http/Controllers/Configuration/AccessController.php
```diff
@@ -108,6 +108,20 @@ public function editRole($role_id)
...
+    {
+        $this->validate(request(), [
```
leonjza: As this is a POST request, please use a request validator (injected as `Request $request` instead of `$role_id`).

src/lang/en/seat.php
leonjza: Yeah global seems weird here. Could you perhaps explain what you intended with this?

herpaderpaldent:
I wonder if using the title part with a edit-button would be suffice too? 

https://github.com/eveseat/web/pull/257/files#diff-23009497eedbea4c16ffca2ebaa05569R5

leonjza:
> I don't particularly like the wasted space and extra row in this view.

I have to agree with this.



> `global` is used here as a general purpose. We're editing a role which is composed of sub-entities and extra-information.

Makes sense. I think we need to consider a new name for this though. Maybe coupled with the name above we can say something simple like `Edit` that pops up a modal to edit the name?

src/Http/Controllers/Configuration/AccessController.php
leonjza: Accidentally resolved?

src/Http/Controllers/Configuration/AccessController.php
leonjza: Accidentally resolved?

warlof:
It will be replaced by ACL revamp

PR #259, opened 2018-12-26
feat(header): Add a caret near profile user drop-down

warlof:
In order to make the user drop-down more identifiable, provide a caret
near to it.

Closes eveseat/seat#496


leonjza:
Do you have a screenshot?

warlof:
The one bellow

2018-12-23_11h22_10

I've also added a small animation which is turning the caret while the menu is opening

PR #261, opened 2018-12-26
fix(research): Return raw value to search property

warlof:
Returning raw value into search property used by datatable to init and
apply filters on data.

Closes eveseat/seat#482

---

A funny one, if the variable is called into the html part of the blade, it's not rendered encoded. But if it's called into script section, it's encoded. I've also surround the raw value by addslashes function in order to escape string containing apostrophe (which is the initial reason of the fix).


warlof:
All tested and non-work https://www.owasp.org/index.php/XSS_Filter_Evasion_Cheat_Sheet

leonjza:
There is a deeper framework level thing at play here. I'll test this a bit later.

PR #275, opened 2019-01-16
Fix corp killmails table header missarrangement

herpaderpaldent:
does what the title says fix https://github.com/eveseat/seat/issues/509

### DIFF
--- a/src/resources/views/corporation/killmails.blade.php
+++ b/src/resources/views/corporation/killmails.blade.php
@@ -17,9 +17,9 @@
         <thead>
         <tr>
           <th>{{ trans('web::seat.date') }}</th>
-          <th>{{ trans('web::seat.victim') }}</th>
-          <th>{{ trans('web::seat.location') }}</th>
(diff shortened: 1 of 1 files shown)


src/resources/views/corporation/killmails.blade.php
```diff
@@ -18,8 +18,8 @@
         <tr>
           <th>{{ trans('web::seat.date') }}</th>
           <th>{{ trans('web::seat.victim') }}</th>
```
warlof: this column must be ship type

src/resources/views/corporation/killmails.blade.php
```diff
@@ -18,8 +18,8 @@
...
-          <th>{{ trans('web::seat.ship_type') }}</th>
           <th>{{ trans('web::seat.location') }}</th>
+          <th>{{ trans('web::seat.ship_type') }}</th>
```
warlof: this column is location

herpaderpaldent:
updated

PR #281, opened 2019-02-03
chore(license): Update license header to cover 2019

warlof:


leonjza:
Automatically updated via StyleCI.

PR #286, opened 2019-02-21
Working db_tests

herpaderpaldent:
* cd into package folder
* use `composer install`
* run `vendor/bin/phpunit`


herpaderpaldent:
image

tests/TestCase.php
```diff
@@ -0,0 +1,75 @@
...
+        $this->withFactories(__DIR__.'/../src/database/factories');
+    }
+    protected function getPackageProviders($app)
+    {
+        // ConsoleServiceProvider required to make migrations work
+        return [
+            WebServiceProvider::class,
+        ];
+    }
+    protected function getPackageAliases($app)
```
warlof: Since this is doing nothing, it has to be drop.

tests/TestCase.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        // For the facade
+        /*return [
+            'ShopifyApp' => \OhMyBrew\ShopifyApp\Facades\ShopifyApp::class,
+        ];*/
+    }
+    protected function resolveApplicationHttpKernel($app)
```
warlof: Since this is doing nothing, it has to be drop.

tests/WebTest.php
```diff
@@ -0,0 +1,57 @@
...
+use Illuminate\Support\Facades\DB;
+
+class WebTest extends TestCase
+{
+    /**
+     * A basic test example.
+     *
+     * @return void
+     */
+    public function testExample()
```
warlof: this must be drop as it's not testing anything.

PR #289, opened 2019-03-02
Feat/async mining ledger

herpaderpaldent:
This enables asyncronous loading of the mining ledger and introduces aggregation of users in group:

requires: eveseat/services#89 eveseat/services#91

image


herpaderpaldent:
Now requires: https://github.com/eveseat/services/pull/91

PR #305, opened 2019-05-24
Update datatables.min.js RowGroup Plugin to Nightly Build

darkflamed:
Edit RowGroup to nightly build for in order to fix datatables errors on null values.

https://datatables.net/forums/discussion/54912/rowgroup-1-1-0-referenceerror-that-is-not-defined-when-datasrc-is-null


warlof:
Hi @darkflamed , thank you for the contribution.
Just for tracking, can you explain quickly which error are fixed ?

leonjza:
What does this fix?

darkflamed:
For example the asset page has an error for characters with a lot of assets from RowGroups, saying  "that is not defined". Found this post about the specific error https://datatables.net/forums/discussion/54912/rowgroup-1-1-0-referenceerror-that-is-not-defined-when-datasrc-is-null and it lead me to the nightly build. Thanks

warlof:
Perfect, will check this tomorrow :)

PR #306, opened 2019-05-26
Fix eveseat/seat#561

wfjsw:


wfjsw:
Would be better if there is an option to enable/disable character grouping. 🤔

warlof:
Thanks for the fix, it's all good for me.
Regarding the "grouping", to be honest, I'm not fan of the thing :p

We probably have to find a better way to show that kind of information sometime.

PR #307, opened 2019-05-29
ACL UI revamp

warlof:
This provide a complete revamp of the Access Control Layer interface in order to make it more user friendly.

It introduce some new role information fields dedicated to third party integration like the description and logo.
Roles can now be renamed which was a lack of feature since the beginning.

The affiliation mechanisms has been renamed into "filters" to be more expressful about what it's doing (restricting access to the entities in the filters).

Filters can now target alliance in addition of the two others (character and corporation).

Permissions has been splitted in scope and include description in order to assist admin while he's building its role.

The interface is also providing a quick "check all" button near each scope which will allow admin to quickly design its roles.

Filters are now stored into a JSON object instead of plain SQL relationship which should be more flexible than prior. The current pattern is designed as bellow but not restricted to :
```
{
   $type: [$entities],
   "character": [{"id": $character_id, "text": $character_name}, ...],
   "corporation": [{"id": $corporation_id, "text": $corporation_name}, ...],
   "alliance": [{"id": $alliance_id, "text": $alliance_name}, ...]
}
```

Regarding roles member management, removal is dynamic and you can remove multiple members without having to refresh the page (using ajax).

Member addition, role meta modification and permission modification are only applied once the complete role is saved.

image


src/Config/web.config.php
```diff
@@ -23,4 +23,38 @@
 return [
     'version'                  => '3.0.20',
     'queue_status_update_time' => 10 * 1000, // milliseconds = seconds x 1,000
+    'esi_roles_map'             => [
```
leonjza: Any reason why this was moved? I prefer not to put static stuff like this in configuration files. We could maybe have a separate map file for this if needed.

src/Config/web.permissions.php
```diff
@@ -25,61 +25,166 @@
...
-        'notifications',
-        'pi',
-        'research',
-        'skills',
-        'standings',
-        'sheet',
-        'journal',
-        'transactions',
+        'asset' => [
+            'label' => 'Grant access to Character Assets',
```
leonjza: Translations :D

warlof: Ah yeah, still on the TODO :D
I was waiting for user feedback, but I could probably send them by the meantime.

src/Config/web.config.php
warlof: It's mostly due to class refactoring and I got the filling it was almost a kind of configuration stuff.
If we're putting this map in another file, in which namespace should it be put ?

leonjza: I suggest `Seat\Web\Acl`.

warlof: addressed with https://github.com/warlof/eveseat-web/blob/acl-ui-revamp/src/Acl/EsiRoleElement.php and https://github.com/warlof/eveseat-web/blob/acl-ui-revamp/src/Acl/EsiRolesMap.php

warlof:
For testing, you will have to alter your `composer.json` file in order to pull the `eveseat/web` package from my own repository. This will probably not be merged until the end of tests. Repository : `https://github.com/warlof/web.git`
Branch : `acl-ui-revamp` (according to composer convention, you must set the version to `dev-acl-ui-revamp`)

You'll get more information here https://getcomposer.org/doc/05-repositories.md#loading-a-package-from-a-vcs-repository
Basically, it's

```
...
"prefer-stable": true,
"repositories": [
  {
    "type": "vcs",
    "url": "https://github.com/warlof/web"
  }
],
"require": {
    [...]
    "eveseat/services": "~3",
    "eveseat/web": "dev-acl-ui-revamp"
}
```

PR #342, opened 2019-12-15
feat(squads): provide intermediate layer between user & ACL

warlof:
This is an initial pull request providing UI and structure for new Squads feature.

Squad should operate as an intermediate layer between user & ACL.
ACL are designed (and should be limited to) to grant or deny access to a resource.

In order to make them easier to apply, squads will act as auto bot, applying selected role to squad member as soon as the user join it.

They also should help to split community virtually, without needing to create empty roles.

TODO :
 - [x] append a filter modal, providing multiple kind of criteria to determine who's able to join a squad
 - [x] add squad creation UI
 - [ ] implement squad logo rendering
 - [x] add squad type (auto or "open") when auto will recruit an user automatically and open will need user to join the squad manually
 - [x] add UI for roles management
 - [x] make roles to automatically be assigned or revoked depending on cases

image

image

image


Crypta-Eve:
Looks great! I have a couple of requests though...

1. A hidden squad type where moderators can add members, but members cannot see or apply to the squad themselves.

2. Have the ability to have the moderators of one squad be determined by their membership status of another squad.

3. Have the ability to have a squad auto remove members based on corp/alliance filters....

Just my 2c

warlof:
@Crypta-Eve auto remove is induce with filters.

Typically, you can define multiple filters, almost like a query builder, composed of different kind of stuff (Corporation Info, Alliance Info, invType, etc...)

About first point, it's noticed. However, does it have any valuable interest compared to auto squad type or manual role assignment? 

Second point sounds difficult. If I implement it, that will not be for initial 4.

Crypta-Eve:
Cool, with the auto remove was just making sure it would be auto remove and not just prevent initial joining.

I think the advantage of point 1 could be something like having a spec ops, invite only SIG. I dont want to give the leaders the ability to modify ACLs, and I dont want the generic line members to know the group exists. Does that make sense as a use case?

Second point, I havent put eyes into how squads work.. But potentially, all members with Role X are automatically made moderators of Squad A. All members of Squad B receive role X. Or am I thinking about this poorly?

warlof:
just started to implement the filter component, rougly inspired from that vuejs stuff https://dabernathy89.github.io/vue-query-builder/

image

If I did it right, I think that would be super easy to extend. For any of your custom filter, build an element composed of :
 - label : the thing which will be displayed on UI (using translation string)
 - src : the route from which data have to be fulled (using select2 format)
 - path : the dotted travellable path from source model up to target searched field

PR #343, opened 2019-12-22
Fix the wrong Language code of Simp. Chinese

wfjsw:
**Breaking Change**

According to ISO 639-1, language code of Chinese should be `zh` instead of `cn`.  
Additionally, `zh-CN` is used to make space for other variant, such as `zh-TW`.

Reference: https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes


warlof:
Hi @wfjsw 

According to your link, why didn't you use short `zh` which meet currently used ISO ?!
Do you plan to include support for country specific languages ?

Which mean we have to switch current `en` to `en-US`, `fr` to `fr-FR`, `ru` to `ru-RU`, `de` to `de-DE`.

I think we can stick to 639-1 without country suffix.

wfjsw:
I'm ok with `zh` only, but people may want to add other regional language. Feedback needed.

warlof:
@wfjsw could you apply your change to other languages ? (switch to IETF language tag format)

Even if we stay on the current implementation of translations, that could open things for ESI (allowing us to exploit `Accept-Language` header - ie: `/universe/types/{type_id}` endpoint)

/ ! \ That change may also imply an update of translation directories.

wfjsw:
Got it.

I have another pull request scheduled which enables automatic language negotiation with browser. It is a stub currently. Can you help me complete it?

warlof:
I didn't figure they were using different formats -_-
Let's go for a l10n version and extending it to i18n format when needed (ie: asian languages).

Laravel standard is working well with it anyway.

| Current | New | ESI variant |
| ------- | ---- | ----------- |
| af | af | unsupported => en-US will be returned |
| cn | zh-CN | zh |
| de | de | de |
| en | en | unsupported => en-US will be returned |
| fr | fr | fr |
| ru | ru | ru |

I'll merge your PR as in tomorrow morning.

warlof:
thanks :)

PR #349, opened 2020-01-07
Update permissions.php

autophob:
Applied possible translation changes to make permissions a bit more understandable.


warlof:
Big thanks for that commit - I hope it will help :)

PR #351, opened 2020-01-11
german translations

ppfeufer:
For the wallet translations I struggled, there are a lot of transaction types I couldn't find a useful translations for. Mainly because I don't know any guys in my corp/alliance who have a german client and all transaction types available in their transaction history.


warlof:
Hi @ppfeufer thanks for that pull request. Can you exclude your modifications from the `.gitignore` file though?

PR #379, opened 2020-02-16
fix(search): Fixes to DataTables errors on Search page

benthompson0:
This fixes the various DataTables errors on the search page, mostly due to the need to update our corporation and alliance relation to be through the affiliation model from recent changes. Also adding a missing language string for faction to the English language file.


warlof:
muchas gracias :)

PR #380, opened 2020-02-18
 fix: AccessCheck affiliation lookup, missing cron validator when adding schedules.

benthompson0:
Fixes PDOException (column corporation_id not found) when using getAffiliationMap() to get effective user permissions for non-superusers.

Fix issue when adding Schedule entry with the Cron validator that was removed in commit c65c62c72feba682a259b2fc19ba5e870f0cb56d. Illuminate\Contracts\Container\BindingResolutionException. Target class [Seat\Web\Http\Validation\Custom\Cron] does not exist.

### DIFF
--- a/src/Acl/AccessChecker.php
+++ b/src/Acl/AccessChecker.php
@@ -265,7 +265,9 @@ trait AccessChecker
         foreach ($corporations as $corporation) {
 
             // for each of them, grant access to all corporation members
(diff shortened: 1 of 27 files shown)


benthompson0:
Added an additional fix to a HTTP 500 during login caused by a SQL duplicate key constraint on the character_id column in RefreshTokens model when logging in as a character that has changed accounts. Will log this in the security log but allow login to complete.

PR #381, opened 2020-02-19
fix(notifications): fix datatables error on character notifications from issue passing notification text from array to string

benthompson0:
The Accessor on the Character Notifications model parses the Yaml stored in the DB, but returns an array. Trying to access this results in a PHP error parsing the array to string. Also updated the Bootstrap 3 popover component to the new Bootstrap 4 tooltip component.

### DIFF
--- a/src/resources/views/character/notification.blade.php
+++ b/src/resources/views/character/notification.blade.php
@@ -1,3 +1,3 @@
-<i class="fa fa-comment" data-widget="popover" data-placement="top" title="" data-html="true"
-   data-trigger="hover" data-content="{{ clean_ccp_html($row->text) }}"></i>
+<i class="fa fa-comment" data-toggle="tooltip" data-placement="top" data-html="true"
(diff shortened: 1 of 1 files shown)


warlof:
What about using `clean_ccp_html($row->getOriginal('text')) ` instead of `clean_ccp_html($row->getAttributes()['text']) ` ?

benthompson0:
I was probably over-thinking it a bit too much - when looking into this I came across [this PR](https://github.com/laravel/framework/pull/31470) against the Laravel Framework which wants to change the behaviour of getOriginal($key) to return the mutated value, which is undesirable in this case as we want a string, not an array returned - and it looks like this was integrated for Laravel 7.x ([commit](https://github.com/laravel/framework/commit/b1c3070fba11d8f47623350287738ed10893c856)), so I initially went with getAttributes()[$key] as it seems the behaviour won't change in the near future. getRawOriginal($key) will replace this to get the original value sans-mutator.

PR #393, opened 2020-03-16
Permission changes

JLCarveth:
Three new Moon Reporter permissions:  
- create_moon_reports
- view_moon_reports
- manage_moon_reports

Changed all references to old permissions to the appropriate(?) new permissions.

### DIFF
--- a/src/Http/DataTables/Tools/MoonsDataTable.php
+++ b/src/Http/DataTables/Tools/MoonsDataTable.php
@@ -23,6 +23,7 @@
 namespace Seat\Web\Http\DataTables\Tools;
 
 use Seat\Eveapi\Models\Sde\MapDenormalize;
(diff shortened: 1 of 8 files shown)


src/Http/DataTables/Tools/MoonsDataTable.php
```diff
@@ -160,7 +161,8 @@ public function html()
      */
     public function query()
     {
-        return MapDenormalize::moons()
+        return MapDenormalize::whereIn(
```
warlof: Can you improve the query using a whereHas statement instead ?
This one will be bounced by engine if there are too many entries in the generated list.

warlof: will sort it after merge, thanks for the contribution ;)

PR #394, opened 2020-03-17
Feature: generate link that provides another SeAT user access to your characters

benthompson0:
Adds a new area to a users profile to generate a link that can be shared with (for example) recruiters to provide them access to any of the users linked characters. 

The link is only valid for users to activate within 7 days of it being generated, it must then be regenerated. Access to the linked characters granted by the sharing link is valid for the duration of the users session (the link can be used any number of times prior to its expiry though).

Closes eveseat/seat#603.


benthompson0:
This PR depends on eveseat/eveapi#202 being merged (uses the updated relationship).

warlof:
Would it be possible to improve a bit the pr to include
 - ability to choose duration
 - log in security event the access link creation
 - log in security event the link usage
 - limit à link to a specific character and not all of them
 - reword sharing link to sharelink

benthompson0:
PR has been updated based on @warlof feedback.

warlof:
thank you very much :)
almost certain it will make happy peeps :p

PR #423, opened 2020-04-23
Auto Join squad if there are no moderators

darkflamed:


warlof:
That would be better to bypass the complete candidate flow. I don't see any utility to ask user to fill a form which will not be used.

benthompson0:
Maybe add a new type of squad that is "open" to all people to opt-in/opt-out of at will. If a group has no moderator, it may not be desirable to automatically allow people to join it in case it applies a role that you don't want just anyone to have.

warlof:
@jpbrowning if you don't put moderator on a group, you don't want it to be moderate. And if you don't want a group beeing visible by other - you're using the hidden group.

What would be the use case to get a manual group, without moderators and without auto opt-in ?

benthompson0:
@warlof never mind, my comment was based on an assumption from an older version where moderators where set as a pivot flag, so removing the member of the group (say the member who was moderator was no longer affiliated to the right corp) would remove the moderator. I see the moderator is set separately so this wouldn't happen - thus my concern isn't relevant anymore.

PR #424, opened 2020-05-02
enhance sso: ability to define scope profiles

benthompson0:
This PR adds support for defining multiple SSO scope profiles in the settings to allow different access levels depending on the type of user joining SeAT - for example different scope profiles for corporation, alliance and public.

Profiles are accessed using the route auth.eve.profile with the parameter of the profile name. Examples:
/auth/eve (for default profile, route via alias: `route('auth.eve')`)
/auth/eve/profile/alliance (for a profile named alliance, route via alias: `route('auth.eve.profile', 'alliance')`)

Default profile: 
image

Custom profile example: 
image 

This may address and close eveseat/seat#382 - although the control over scopes stays with the SeAT administrator rather then the user, this allows extra flexibility that I think addresses the reason for requesting this capability.

### DIFF
--- a/src/Http/Controllers/Auth/SsoController.php
+++ b/src/Http/Controllers/Auth/SsoController.php
@@ -63,7 +63,7 @@ class SsoController extends Controller
         } else {
             // Get the scopes that are marked as the default.
             $scopes = $scopes_setting->first(function ($item) {
-                return $item->name == "default";
+                return $item->name == 'default';
             });
         }
 
--- a/src/Http/Controllers/Auth/SsoController.php
+++ b/src/Http/Controllers/Auth/SsoController.php
@@ -43,14 +43,36 @@ class SsoController extends Controller
      * Redirect the user to the Eve Online authentication page.
      *
      * @param \Laravel\Socialite\Contracts\Factory $social
+     * @param string $profile
      * @return mixed
      * @throws \Seat\Services\Exceptions\SettingException
      */
-    public function redirectToProvider(Socialite $social)
+    public function redirectToProvider($profile = null, Socialite $social)
     {
 
+        $scopes_setting = collect(setting('sso_scopes', true));
+
+        if(! is_null($profile)) {
+            $scopes = $scopes_setting->first(function ($item) use ($profile) {
+                return $item->name == $profile;
+            });
+
+            // Invalid profile name?
+            if(is_null($scopes))
+                abort(400);
+        } else {
+            // Get the scopes that are marked as the default.
+            $scopes = $scopes_setting->first(function ($item) {
+                return $item->name == "default";
+            });
+        }
+
+        // Store the scopes we are sending to CCP in the session so we can
+        // validate the JWT response contains the right scopes.
+        session()->put('scopes', $scopes->scopes);
+
         return $social->driver('eveonline')
-            ->scopes(setting('sso_scopes', true))
+            ->scopes($scopes->scopes)
             ->redirect();
     }
 
--- a/src/Http/Controllers/Auth/SsoController.php
+++ b/src/Http/Controllers/Auth/SsoController.php
@@ -54,7 +54,7 @@ class SsoController extends Controller
 
         // Get the scopes that are marked as the default.
         $scopes = $scopes_setting->first(function ($item) {
-            return $item->name == 'default';
+            return $item->default == true;
         });
 
         if(! is_null($profile)) {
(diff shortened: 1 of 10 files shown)


src/Http/Controllers/Auth/SsoController.php
```diff
@@ -43,14 +43,36 @@ class SsoController extends Controller
...
      */
-    public function redirectToProvider(Socialite $social)
+    public function redirectToProvider($profile = null, Socialite $social)
     {
 
+        $scopes_setting = collect(setting('sso_scopes', true));
+
+        // Get the scopes that are marked as the default.
+        $scopes = $scopes_setting->first(function ($item) {
+            return $item->name == 'default';
```
warlof, line 57: You add a `default` property to your object, why not using it instead the name ?
I think it would be more reliant ;)

And doing it so will allow you to determine which scope-profile is the default one.

src/Http/Controllers/Configuration/SsoController.php
```diff
@@ -34,10 +35,22 @@ class SsoController extends Controller
...
      */
-    public function getConfigurationHome()
+    public function getConfigurationHome(Request $request)
     {
 
-        return view('web::configuration.sso.view');
+        $sso_scopes = collect(setting('sso_scopes', true));
+
+        $selected_profile = $sso_scopes->first(function ($item) {
+            return $item->name == 'default';
```
warlof, line 44: You add a `default` property to your object, why not using it instead the name ?
I think it would be more reliant ;)

src/Http/Controllers/Configuration/SsoController.php
```diff
@@ -49,7 +62,26 @@ public function getConfigurationHome()
     public function postUpdateScopes(SsoScopes $request)
     {
 
-        setting(['sso_scopes', $request->input('scopes')], true);
+        $scopes = collect(setting('sso_scopes', true));
+
+        // default profile cannot be renamed
+        $default_profile = $scopes->first(function ($item) {
+            return $item->name == 'default';
```
warlof, line 69: You add a `default` property to your object, why not using it instead the name ?
I think it would be more reliant ;)

src/Http/Controllers/Configuration/SsoController.php
```diff
@@ -49,7 +62,26 @@ public function getConfigurationHome()
...
+        });
+
+        if($default_profile->id == $request->input('profile_id') && $request->input('profile_name') != 'default')
```
warlof, line 72: You add a `default` property to your object, why not using it instead the name ?
I think it would be more reliant ;)

src/Http/Controllers/Configuration/SsoController.php
```diff
@@ -70,10 +112,26 @@ public function getEnableAll()
...
-    public function getRemoveAll()
+    public function getDeleteProfile(int $id)
     {
 
-        setting(['sso_scopes', []], true);
+        $scopes = collect(setting('sso_scopes', true));
+
+        // default profile cannot be removed
+        $default_profile = $scopes->first(function ($item) {
+            return $item->name == 'default';
```
warlof, line 122: You add a `default` property to your object, why not using it instead the name ?
I think it would be more reliant ;)

benthompson0:
Updated to use default attribute.

benthompson0:
Added a new commit containing functionality to allow customisation of the sign in page and adding login buttons using the various scope profiles.

warlof:
The overall idea is good, I like to be able to customize multiple profile, giving them a name and select a default profile.

Some quick feedback regarding changes :
 - instead showing an HTTP 400 if the provided profile is wrong, just use the default one ?
 - the SSO profile button placeholder will need to be improved as not only valid ones are replaced
 - I was more expecting a simple drill-down to determine which profile to be use instead multiple sign-in buttons. Current version may generate a really big page (assuming you got lot of profile).

Regarding placeholders, I added an second profile called "Alliance" to the default existing one.
I then fill a few things in the Quill editor, using the generated placeholders and adding extra random stuff in brackets. As a result, all stuff which was meeting the pattern has been replaced. It should only replace existing profile (`[[Alliance]]` and `[[default]]` in bellow example)

```
You are part of the Block : [[Alliance]]

In every other case : [[default]]

[[bugtrigger]]
```

Also, you still have commented block from current version :p
https://github.com/eveseat/web/pull/424/files#diff-76070a4b45d53742e0fe21548007c304R18-R23

PR #426, opened 2020-05-03
Switch ACL to standard Gates

warlof:
This pull request is switching all ACL related stuff to standard Laravel [Gates and Policies](https://laravel.com/docs/6.x/authorization)

**Breaking Changes** are related to usage of `auth()->user()->has()` which can be replaced by all Gates methods, including blades `@can`.

To make change easier, 3 dynamic policies has been designed to handle permissions :
 - `Seat\Web\Acl\Policies\CharacterPolicy` will be used by default for all `character` scope permissions
 - `Seat\Web\Acl\Policies\CorporationPolicy` will be used by default for all `corporation` scope permissions
 - `Seat\Web\Acl\Policies\GlobalPolicy` will be used for everything else

You can bypass this flow by providing your own Gate handler in permissions definition using property `gate`. **DO NOT** register your gates, they will be registered by the web package.

The new permission configuration file manifest is described in the table bellow

| Property     | Type                                              | Purpose                                               | Mandatory |
| ----------- | ----------------------------------- | --------------------------------------- | ---- |
| **label**           | translation token                           | Permission display name in the role UI | **yes** |
| **description** | translation token                           | Permission description in the role UI    | no |
| **division**      | assets, financial, industrial, military | UI highlight with color and icon           | no |
| **gate**           | method calling path                      | Used to specify a custom handler for the permission | no |

Also, permissions are cached for **10 seconds** at first Gate call in order to reduce check flow and needed queries during page rendering.

Finally, Squad applications flow has been simplified for type `manual` without moderators. It's a simple opt-in (eveseat/web#423).

Last but not least, ~~`superuser`~~ `global.superuser` has been deprecated and moved to `users` as a simple `admin` flag. This will reduce complexity with roles management and system user `admin`


warlof:
Keep it open for a week waiting feedback

[ ] update ACL tests to relay on Policies instead of `AccessChecker` and `AccessManager`

PR #429, opened 2020-05-17
Moon Reporter Changes

JLCarveth:
- Fixed column ordering 
- Implemented filter based on region, constellation, system, and moon contents


warlof:
replaced by eveseat/web#430

PR #430, opened 2020-05-21
Moon Reporter Changes

JLCarveth:
As per slack:
- Split `MoonScope` into `RegionScope`, `ConstellationScope`, `SystemScope`, and `MoonContentScope` 
- Removed the null checking needed in MoonsController by removing type specificity in scope constructors
- Middleware path should be fixed.
- StyleCI changes should be applied
- Fixed issues with missing reprocessing and price code I believe (git issue)


warlof:
thank you very much to help SeAT being better :)

PR #448, opened 2020-06-29
Adding production environment to horizon config

lk77:
Hello,

i propose to add production environment to horizon config.

currently, workers are not running with APP_ENV=production

thanks.


warlof:
thank you for the change

stuff have been improve a bit with next gen' https://github.com/eveseat/web/blob/4.0.x/src/WebServiceProvider.php#L342-L346
so I think it doesn't need an update for 4.0.0

PR #455, opened 2020-07-25
fix(users): improve consistency regarding user migration

warlof:


src/database/migrations/2019_11_12_220840_drop_groups_table.php
```diff
@@ -52,11 +52,28 @@ public function up()
                 ->delete();
         }
 
+        // drop orphan groups
+        DB::table('groups')
+            ->whereNotIn('id', DB::table('users')->slect('group_id'))
```
Crypta-Eve, line 57:
```suggestion
            ->whereNotIn('id', DB::table('users')->select('group_id'))
```

PR #456, opened 2020-07-31
fix(translations): corporation mining ledger datatable user header

Crypta-Eve:
Fix the header on the corp mining ledger data table to correctly be translated


Crypta-Eve:
Closing this as more needs to be done first

PR #458, opened 2020-08-04
fix(mail): Mail Timeline and Character Mail

Crypta-Eve:
- character mail timeline now functional and respects any given roles
- character mail view fixed for orphaned users


warlof:
thank you for this fix :)

PR #468, opened 2020-08-18
Adding corp titles as a squad filter

d3vfreak:
Adding corp titles as a squad filter so you can auto assign groups to directors for example.


warlof:
thank you :)

PR #492, opened 2020-09-27
Refactor total SP stats to use all associated character IDs on homepage

Crecket:
The getTotalSP method is only used on the homepage so I assume it is safe to refactor like this. 

Related to eveseat/seat#673


warlof:
Hi,

Thank you for this pull request. At pure code point of view, it seems correct.
I changed that metric with SeAT 4 release as I got the feeling suming SP accross all toons wasn't really accurate.

warlof:
thank you for this enhancement :)

PR #500, opened 2020-11-01
feat(squads): add ability to restrict by skill level

warlof:
revamp the filter pattern to allow to pair rules by relation paths.
this allow to define a filter based on both skill name and skill level.

Closes eveseat/seat#693

### DIFF
--- a/src/Models/Filterable.php
+++ b/src/Models/Filterable.php
@@ -22,6 +22,7 @@
 
 namespace Seat\Web\Models;
 
+use Illuminate\Database\Eloquent\Builder;
 use Illuminate\Database\Eloquent\Model;
 use stdClass;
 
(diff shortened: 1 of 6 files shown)


TaintedSoulz:
While you are editing the filterable class we have a huge problem, with the query you are using you are comparing all of the characters the user has at once per rule, so for instance:

Squad:
  Rules:
    Character has to be in alliance XYZ
    Character has to have asset Titan

When the rule gets looked over and applied, the same character does not have to have to be in the Alliance, and also have a titan. Any character can have a titan, and any character can be part of the alliance, and it will still return true. The character that the rule is being applied to and tested on should have the following effect: (single character in alliance XYZ AND has a Titan)

PR #509, opened 2020-11-29
Allow CEO and Directors to get a list of characters and corporations

TaintedSoulz:
Also allow directors to access the characters pages of the corporation and the corporation page itself.

### DIFF
--- a/src/Acl/Policies/CharacterPolicy.php
+++ b/src/Acl/Policies/CharacterPolicy.php
@@ -125,13 +125,10 @@ class CharacterPolicy extends AbstractEntityPolicy
         }
 
         // attempt to retrieve roles for any associated character related to the corporation
-        $corporationRoles = CorporationRole::where('corporation_id', $corporation_id)
+        return CorporationRole::where('corporation_id', $corporation_id)
             ->whereIn('character_id', $user->associatedCharacterIds())
             ->where('role', 'Director')
-            ->where('type', 'roles')->get();
-
-        // if user has any director roles in the corporation return true
-        return $corporationRoles->count() > 0;
+            ->where('type', 'roles')->exists();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


TaintedSoulz:
Looks like to pass the continuous integration the database needs to have the table corporation_roles, not sure why its saying this table does not exist.

warlof:
The corporation role table isn't part of the test suite. https://github.com/eveseat/web/tree/master/tests/database/migrations

Can't use character roles instead?

TaintedSoulz:
The problem is if i'm using the character roles in this case it could come out of sync. So for instance a users affiliation could change, and then the roles would need to be updated, and if they weren't the director role could transfer over to a different corporation. It just makes it so the check is more fluid and not relying on more then one updated piece of data.

src/Acl/Policies/CharacterPolicy.php
```diff
@@ -107,6 +109,31 @@ private function isCeo(User $user, CharacterInfo $character)
...
+        // in case we were not able to find the corporation ID - assume the user is not Director of their corporation
+        if (is_null($corporation_id)) {
+            return false;
+        }
+
+        // attempt to retrieve roles for any associated character related to the corporation
+        $corporationRoles = CorporationRole::where('corporation_id', $corporation_id)
+            ->whereIn('character_id', $user->associatedCharacterIds())
+            ->where('role', 'Director')
+            ->where('type', 'roles')->get();
```
warlof, line 131: why do you need to get here ? you just want to know if you have a match. You can use `exists`

src/Acl/Policies/CorporationPolicy.php
```diff
@@ -96,6 +97,22 @@ private function isCeo(User $user, CorporationInfo $corporation)
...
+     * @param \Seat\Web\Models\User $user
+     * @param \Seat\Eveapi\Models\Character\CharacterInfo $character
+     * @return bool
+     */
+    private function isDirector(User $user, CorporationInfo $corporation)
+    {
+        $corporationRoles = CorporationRole::where('corporation_id', $corporation->corporation_id)
+            ->whereIn('character_id', $user->associatedCharacterIds())
+            ->where('role', 'Director')
+            ->where('type', 'roles')->get();
```
warlof, line 110: why do you need to get here ? you just want to know if you have a match. You can use `exists`

PR #510, opened 2020-11-30
Allow roles to be used as filters for squads

TaintedSoulz:


src/Http/Controllers/Support/FastLookupController.php
```diff
@@ -526,4 +527,29 @@ public function getSystems(Request $request)
             'results' => $systems,
         ]);
     }
+
+    /**
+     * @param \Illuminate\Http\Request $request
+     * @return \Illuminate\Http\JsonResponse
+     */
+    public function getRoles(Request $request)
```
warlof, line 535: Wouldn't it be better to provide a nicely built corporation roles list ?
This will require that all roles exists in table to be used.

Alternatively, use a `distinct` statement rather than a group by one since you just want the role name minus all its duplication. The `distinct` statement will be faster than the group one as it can relay on the index directly where the `groupBy` is using result buffer :)

Also, can we rename `getRoles` to `getCorporationRoles` ? Just to avoid any confusion with ACL.

TaintedSoulz: My thinking was to leave it open for possible future roles, I have a bad habit of really disliking hard coding things that could possibly change. If you are really wanting it changed past just a glance just let me know. For now I implemented distinct.

PR #550, opened 2021-01-31
fix: migration fails due to malformed foreign key constraint in MariaDB

benthompson0:
MariaDB doesn't accept a foreign key constraint unless the columns on both sides are the same type, including if one side is an index. The moon_id column in universe_moon_contents is an index, so the moon_id column in universe_moon_reports must be as well in order to have a foreign key constraint between the two.

Ref: https://blog.mariadb.org/mariadb-innodb-foreign-key-constraint-errors/


warlof:
Hum, `moon_id` must be a primary key here. I don't know why it's not the case.

benthompson0:
I've updated the change to a primary key and rerun the migration.

PR #551, opened 2021-02-08
fix: Replace today's date with a string when no remap exists.

maxtsero:
Currently a char's remap dates are shown as today if the value is `null` this PR replaces today's date with `No Remap Found` in those cases.

### DIFF
--- a/src/resources/views/character/sheet.blade.php
+++ b/src/resources/views/character/sheet.blade.php
@@ -198,9 +198,9 @@
                   <dt>{{ trans('web::seat.bonus_remaps') }}</dt>
                   <dd>{{ $character->pilot_attributes->bonus_remaps ?: 0 }}</dd>
                   <dt>{{ trans('web::seat.last_remap_date') }}</dt>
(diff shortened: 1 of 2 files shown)


src/resources/views/character/sheet.blade.php
```diff
@@ -198,9 +198,9 @@
                   <dt>{{ trans('web::seat.bonus_remaps') }}</dt>
                   <dd>{{ $character->pilot_attributes->bonus_remaps ?: 0 }}</dd>
                   <dt>{{ trans('web::seat.last_remap_date') }}</dt>
-                  <dd>{{ $character->pilot_attributes->last_remap_date ?: carbon() }}</dd>
+                  <dd>{{ $character->pilot_attributes->last_remap_date ?: trans('web::seat.no_remap') }}</dd>
                   <dt>{{ trans('web::seat.accrued_remap_cooldown_date') }}</dt>
-                  <dd>{{ $character->pilot_attributes->accrued_remap_cooldown_date ?: carbon() }}</dd>
```
warlof, line 203: isn't that still valid if there is no remap ? I think it was the "next available remap date"

warlof:
Thank you very much

PR #568, opened 2021-04-23
Open links to horizon queue in a new tab

Nfinished:
Never in my life have I meant to navigate away from seat when visiting horizon


warlof:
Hi,

Thank you for the change 🙂

PR #569, opened 2021-04-24
Confirm destructive operations

maxtsero:
This PR adds a new `bootbox` confirmation dialog for deletion and uses the existing `confirmform` dialog so that (I hope?) all destructive operations have confirmation either in the form of `Are you sure you want to delete ${entity}` or in the form of `Are you sure you want to continue?`


warlof:
oh, I think it was merged - thank for the update, well done :)

PR #570, opened 2021-05-14
Add CN translation

ArielHeleneto:
Add CN translation


warlof (commented):
Hi,

Some captions use singular/plural notation and seems not beeing translated accordingly. Is that intended? (especially in the moon file)

Also, there are remaining English caption. Is it intended as well?

ArielHeleneto:
> Hi,
> 
> Some captions use singular/plural notation and seems not beeing translated accordingly. Is that intended? (especially in the moon file)
> 
> Also, there are remaining English caption. Is it intended as well?

For the first question , of course yes. In Chinese , we don't have regular form.It is the same. I have tried in my SeAT system, it works as Chinese's minds.

second question, some of them are intended because we are still having conversations. Eve do not have Chinese official version. We do keep it as same as it performed in the game.

ArielHeleneto:
src/resources/lang/zh-CN/seat.php and src/resources/lang/zh-CN/permissions.php are not translated yet because i don't have much time

tehraven:
@ArielHeleneto this is a huge step forward for Chinese players using SeAT. Are you able to find another Chinese speaker who can finish up? This would be wonderful to get merged!

warlof:
if you remove the untranslated entries, it's mergeable.
we don't need all keys to be translated - when a caption is not found in target language, the default one is selected (english)

ArielHeleneto:
> @ArielHeleneto this is a huge step forward for Chinese players using SeAT. Are you able to find another Chinese speaker who can finish up? This would be wonderful to get merged!

emmm,i can not find another one to help me.

PR #572, opened 2021-05-19
Implemented filtering based on current selection

NickvdMeij:
The current filter option worked lightly different from what users expect to see. Instead of filtering on a range of moons, it filtered only moons that has all the selected materials, instead of showing a compromised list that the user wanted.

For example, say you want to only see R16,R32 and R64 moons. With the previous setup, this was simply not possible since you would end up with a query like

->andWhere(moon is R16)->andWhere(moon is R32)->andWhere(moon is R64)

Altering the format of the query allows for all the same usecases at it previously provided, as well as support other usecases.

### DIFF
--- a/src/Http/DataTables/Scopes/Filters/MoonRankScope.php
+++ b/src/Http/DataTables/Scopes/Filters/MoonRankScope.php
@@ -54,7 +54,7 @@ class MoonRankScope implements DataTableScope
      **/
     public function apply($query) {
         $query->whereHas('content', function ($type) use ($group_id) {
-            $type->->whereIn('marketGroupID', $this->groups);
+            $type->whereIn('marketGroupID', $this->groups);
         });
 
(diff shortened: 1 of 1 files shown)


warlof:
@NickvdMeij this was the expected flow, but I may figure why people would be able to get result of a large selection rather than meeting exactly the selection.

Can you apply StyleCI change request though ? you're refering to a variable which is not used anymore in your callback :)

NickvdMeij:
> @NickvdMeij this was the expected flow, but I may figure why people would be able to get result of a large selection rather than meeting exactly the selection.
> 
> Can you apply StyleCI change request though ? you're refering to a variable which is not used anymore in your callback :)

Good spot! editted in browser since i dont have a local setup atm haha. Gotta get an IDE setup soon ;P, anyways change is commited and tests are running 👍

PR #575, opened 2021-08-10
Fix backdrop for radar chart ticks

veteranmina:
Corrects issue when using Jet skin, ticks have white backdrop enabled by default.


veteranmina:
Apparently renaming a branch is considered deleting a branch...

PR #578, opened 2021-08-19
Correct form for confirmform

veteranmina:
Corrects form not posting action after confirming action

### DIFF
--- a/src/resources/js/seat.js
+++ b/src/resources/js/seat.js
@@ -10,7 +10,11 @@ $.ajaxSetup({
 // Make your submit button part of class confirmform, and viola
 var currentForm;
 $(document).on("click", ".confirmform", function (e) {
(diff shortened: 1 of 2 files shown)


warlof:
What's the change?

veteranmina:
I moved the contents of <button></button> to inside the form as it does not register the action outside the form.

veteranmina:
To be more specific, this corrects the reject button for squad candidates not functioning as intended. The confirmation form would present itself but confirming the action did not complete due to the button existing outside the form code.

warlof:
`form` attribute is normally designed for such purpose, will need a bit extra investigation

veteranmina:
Reverted reject.blade.php and changed seat.js to review form/attr. This brings compatibility between forms that have buttons existing inside or outside their *.blade.php forms

PR #582, opened 2021-08-28
Add affiliations update button

veteranmina:
Change would add button in summary card to look like this. If you believe a better location would be rather elsewhere let me know 😄 
image

### DIFF
--- a/src/Config/web.jobnames.php
+++ b/src/Config/web.jobnames.php
@@ -24,6 +24,7 @@
 return [
 
     'character' => [
+        'affiliations'     => \Seat\Eveapi\Jobs\Esi\Update\Affiliations::class,
         'assets'           => \Seat\Eveapi\Jobs\Assets\Character\Assets::class,
         'blueprints'       => \Seat\Eveapi\Jobs\Character\Blueprints::class,
         'calendar'         => [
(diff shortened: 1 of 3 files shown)


warlof:
What about putting next to the user button? So we're keeping controls all together.

veteranmina:
This is how it would look moving it down to the footer.
image

veteranmina:
Further work is needed on this. Do not push this to master.

warlof:
As previsouly explained, button is containg "setup" - so when you hit it, it will know what to do and which date to send.
https://github.com/eveseat/web/blob/5907741fbb7c2b9402830d397bd11a95eec3b3b1/src/resources/views/components/jobs/buttons/update.blade.php#L1

When you click on it, the following ajax payload is sent against SeAT

```json
{
    "type": "character",
   "entity_id": 0,
   "job_name" : "Job\FQCN"
}
```

https://github.com/eveseat/web/blob/5907741fbb7c2b9402830d397bd11a95eec3b3b1/src/resources/views/components/jobs/buttons/update.blade.php#L1

Which is processed by this backend https://github.com/eveseat/web/blob/5907741fbb7c2b9402830d397bd11a95eec3b3b1/src/Http/Controllers/Tools/JobController.php#L55

You will need to put a `btn-xs` class to the button if you want to put it in the card footer, otherwise you'll explode the box.

Also, it might need a bit refactor to be able to use affiliation job since as you noticed, all currently used jobs only took a character ID together with a token as parameter - however, the affiliation job is a bucket and will need an array of stuff.

warlof:
Hey @veteranmina , what's the state of this PR

PR #588, opened 2021-12-20
Correct path for CharacterWallet Journal class

jacben13:
Was testing with the updated home page changes and found this bug.


warlof:
thank you very much :)

PR #589, opened 2022-01-03
fix: update moon chunk quantity

Crypta-Eve:
Together with eveseat/eveapi#320 this fixes the moon mining volumes in line with the updated 40k m3 base drilling volume.


warlof:
Thank you crypta 👍

PR #594, opened 2022-01-07
Add download buttons for token tracking

recursivetree:
Allows you to download the list at `/corporations/{id}/tracking`


warlof:
This is due to the datatable factory, we may get more luck with a repository class - but that's another topic.

However, you will need the exact model related to active entry - therefore, you must call `CorporationMemberTracking::with('character', 'refresh_token', 'ship')->find($serialized_character['id']);` into your `collection` override.

Also, since no modification is done after the collection changes, you may use `return` directly instead of initializing a dummy `$collection` variable.

Regarding `heading`, the returned array don't need to be tied to model properties name. It only need to be in the same order than the collection element. So in case you think it would be better to have more human readable things or even translated string, be free to do it :)

As an example, this will still work :

```
    public function headings(): array {
        return [
            'Token status',
            'Name',
            'Location',
            'Current ship',
            'Joined',
            'Last Login',
        ];
    }
```

PR #598, opened 2022-02-20
feat: implement dynamic dashboard

warlof:
The overall purpose of this is to allow third party to provide their own dashboard using plugin system.

The forkflow would be as bellow :
- when we loading homepage, we search for "prefered" dashboard in user settings
- in case user had select a board, we load that one
- otherwise, we will load a default one from core

A drill-down is available upper boards allowing user to change active board with ease.

The underlined idea is to design more contextual dashboard - so user will see only metrics related to current context (ie: admin dashboard will render security, version, load, etc... - whereas character will render character metrics, skills, etc...)

image


recursivetree:
Instead of changing the complete dashboard, maybe allow the users to customize their widgets individually? But otherwise a great idea, maybe something similar is expandable to e.g. the search function?

warlof:
leaving end user the ability to design its own dashboard is definitely not the same complexity level when talking about implementing. I take the feedback though.

Hadn't played with search yet - but the new template used by SeAT 5 is providing some cool toys regarding searchable stuff :)

PR #604, opened 2022-06-15
Add action description for form confirmation modal

recursivetree:
Add action description for form confirmation modal


warlof:
Thank you for your contribution.

Can you init your message variable with non action label ? That will allow the else removal 🙃

recursivetree:
Is this what you want?

PR #623, opened 2022-08-23
ui: admin avatar should be seat logo

Crypta-Eve:
Admin avatar shows as broken as there is no character portrait for character ID 0. Now for admin we use the SeAT logo.

### DIFF
--- a/src/resources/views/includes/header.blade.php
+++ b/src/resources/views/includes/header.blade.php
@@ -36,7 +36,11 @@
       <!-- user-card -->
       <div class="nav-item dropdown">
         <a href="#" class="nav-link d-flex lh-1 text-reset p-0" data-bs-toggle="dropdown" aria-label="Open user menu">
+          @if(auth()->user()->name != 'admin')
           {!! img('characters', 'portrait', auth()->user()->main_character_id, 32, ['class' => 'avatar avatar-sm'], false) !!}
+          @else
+          <img src="{{ asset('web/img/logo.png') }}" width="32px" height="32px" alt="{{ config('app.name') }}" />
+          @endif
           <div class="d-xl-block ps-2">
             <div>{{ auth()->user()->name }}</div>
             <div class="mt-1 small text-muted">{{ auth()->user()->characters->count() }} characters</div>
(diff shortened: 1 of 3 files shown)


warlof:
> I think we should add a user partial like `src/resources/views/partials/character.blade.php` as well

Yeah, that's a good idea. We will probably need to provide three variants for both entity and user block when used with image.

Bootstrap 5 had reduce a lot of duplicated classes and there are various way avatar is used in tablet which might be cool to have 😀

recursivetree:
I'd include the name too, just like the character partial. If the icon needs to be separate for what you originally wanted to do, split it up even further into logo and logo+name

Crypta-Eve:
Hmm, will have a think on that... Top of my head there is only two places that this partial would be used and the format is different... There is the header which looks like 
image

Then there is the user datatable which looks like below
image

I am wondering if I made the partial look like the header, and then the datatable would also inherit the number of characters (previously not shown on the datatable at all)

recursivetree:
Just because it isn't used in the core doesn't mean it won't be used in plugins. For example, both seat-info and seat-alliance-industry have a use-case for it. Right know, both have their own implementation. Having it in a partial makes changes like the seat logo for admins appear consistently throughout seat. Additionally, we have equivalent components for characters, corporations and alliances, so I think users would make sense.

To keep the difference between header and datatable:

one file for the image and one for image+name, where image+name includes the image only file

Crypta-Eve:
Oh sorry didn't mean it wouldn't be used elsewhere, only I can't load a plugin in v5 atm... I haven't even got my plugins displaying correctly.

I have a few ideas and will have a play around with it tomorrow to see how it turns out. But I'm thinking you're right about two partials, though I think I also should paramterise (with reasonable default value) the size of the img tag. That way it's easier to fit in more places.

Crypta-Eve:
Should be good to go now :) 

two partials added. 
header partial uses the new `user-icon` partial to produce 
image

While the new user `partial` gives a link to the main character sheet of the user. (users datatable not being updated due to future change proposal)
image

PR #627, opened 2022-09-06
Settings Page Update 6.0.x

Crypta-Eve:
This PR updates the SeAT Settings page to use new Tabler classes and also remove the dependency on jQuery.

image


Crypta-Eve:
I'd love to have thoughts on the new layout I added here. It's called 6-6_12, based on an even left/right side and a full-width bottom section. I felt it fits here as it makes it a lot easier to work with custom links due to form widths.

The only other concern I have is regarding the table within a card... I am not sure how to make that fill the card content.

warlof:
Hey, as of table, you can check within the character list. There is a class to add to either the body or the table (render call)

Crypta-Eve:
Ok, this is what I get using the `card-table` class. Thoughts?
image

warlof:
Maybe we should keep settings as clean as possible. Should we extract the custom link pane into a dedicated page ?

recursivetree:
We could, although I don't think it really matters. On another note, I have been considering to add the option to link seat-info articles the way custom links work. Maybe we could store the links in a db table instead of a string in the settings

PR #628, opened 2022-09-08
feat: add links to my corp and alliance

Crypta-Eve:
This adds sidebar links to corp and alliance menus that take the user to the appropriate corp/alliance as determined by their main character.

For most users, this saves the extra click through the All Corps/Alliances tab.


warlof:
Nice add !
Don't you have access to the entity while being on affiliation relationship though ?

ie : ->affiliation->corporation

PR #635, opened 2022-11-14
Allow command arguments in schedule

recursivetree:
This pr allows you to add a command with argument to the seat command schedule.
fixes https://github.com/eveseat/seat/issues/867.

### DIFF
--- a/src/Http/Validation/Custom/ArtisanCommand.php
+++ b/src/Http/Validation/Custom/ArtisanCommand.php
@@ -1,0 +1,81 @@
+<?php
+
+/*
+ * This file is part of SeAT
+ *
+ * Copyright (C) 2015 to 2022 Leon Jacobs
+ *
+ * This program is free software; you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation; either version 2 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License along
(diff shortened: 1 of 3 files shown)


leonjza:
How do we prevent OS command injections here without validating the arguments? It's the primary reason I never added support for it.

recursivetree:
these aren't os level commands, these are just artisan commands. And second, if you have access to use the endpoint, you are an admin. In this case, you can probably access the OS anyways

leonjza:
Being an administrator on SeAT does not (and should not) mean you are an OS administrator.

A quick dig again through some of the internals of `\Illuminate\Console\Application` and `\Symfony\Component\Console\Application` to see how commands are dispatched has me think that if we have unfiltered arguments, adding something like `;id` or similar to a command as argument would result in shell expansion getting you arbitrary code execution.

I'm curious enough to PoC this soon actually :)

leonjza:
Alright I spent a few minutes on this and a face value it seems like I am wrong. Wrote a simple command like this:

```php
// app/Console/Commands/ArgPoc.php
<?php

namespace App\Console\Commands;

use Illuminate\Console\Command;
use Illuminate\Support\Facades\Log;

class ArgPoc extends Command
{
    /**
     * The name and signature of the console command.
     *
     * @var string
     */
    protected $signature = 'poc:args {argname}';

    /**
     * The console command description.
     *
     * @var string
     */
    protected $description = 'Argument Injection POC';

    /**
     * Execute the console command.
     *
     * @return int
     */
    public function handle()
    {
	Log::info('running command with args: ' . $this->argument('argname'));

        return Command::SUCCESS;
    }
}
```

And a simple route like this:

```php
// routes/web.php
Route::get('/poc/{args}', function($args) {
    $exitCode = Artisan::call('poc:args ' . $args);
    return $exitCode;
});
```

Injecting a command via the URL as a test seems to have no effect :)

```
http://172.16.182.175:8000/poc/`whoami`
```

recursivetree:
I updated this PR for seat 5

PR #637, opened 2022-12-13
feat: Add character title to character summary

RazeSoldier:


src/resources/views/character/includes/summary.blade.php
```diff
@@ -122,6 +122,10 @@
           </span>
         @endif
       </dd>
+      @isset ($character->title)
```
warlof, line 125:
```suggestion
      @if($character->title)
```

warlof: just for syntax and consistency :)

src/resources/views/character/includes/summary.blade.php
```diff
@@ -122,6 +122,10 @@
           </span>
         @endif
       </dd>
+      @if (! is_null($character->title))
```
warlof, line 125:
```suggestion
      @if(! is_null($character->title))
```

PR #647, opened 2023-04-24
Make it possible to add more skins

recursivetree:
Makes skins a config that can be extended by plugins using `mergeConfigFrom`. Additionally it removes a reference to an old, no longer existing skin.


recursivetree:
can now be merged together with https://github.com/eveseat/services/pull/161

PR #657, opened 2023-11-02
Show alt characters in character summary when token has expired

recursivetree:
fixes https://github.com/eveseat/seat/issues/856

It is currently broken since character with expired tokens have their `refresh_token` soft-deleted. This means currently SeAT can't find the user from a character with expired token since `refresh_tokens` links characters to a user. The fix is to also consider deleted tokens when searching the user from a character.

The `global.invalid_tokens` permission is respected since it only affects `user->alts` and not the first part of the chain, `character->user`

### DIFF
--- a/src/resources/views/character/includes/summary.blade.php
+++ b/src/resources/views/character/includes/summary.blade.php
@@ -27,9 +27,9 @@
     </p>
 
     <ul class="list-group list-group-unbordered mb-3">
-      @if(! is_null($character->refresh_token))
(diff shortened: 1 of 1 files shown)


recursivetree:
I'm pretty sure the CI failure is unrelated to this, since this PR doesn't touch code that failed

Crypta-Eve:
Not sure what happened there, completed successfully when I requested it rerun.

PR #658, opened 2023-11-03
Rewrite Filterable trait used for Squad filters

recursivetree:
## Problem:
@Bluedagger on discord noticed that if a character is not in any alliance, and the squad contains a filter for `alliance is not x`, the user doesn't get added to the squad, even though he should.

## How it has been fixed
While studying the `Filterable` trait code (https://github.com/eveseat/web/blob/c97b7cd1dd8cd5c063415ba0710d4b343dfdfef5/src/Models/Filterable.php), I noticed how messy and broken it was. I really struggled to first pin down the problem, and fixing it was even harder since it would have required to special case `is not` in multiple cases.

Therefore, I decided it's rewrite time! This allowed me to fix the original bug, and additionally:
- Make it more readable: There is no 8 levels deep nested code anymore
- Allow nesting of groups to arbitrary depths
- Add a new test case covering the bug

All squad tests are passing.

Lastly, I added a few files to the gitignore that were generated while running tests, but don't belong on github.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .env.php
 .env
 .idea/
+vendor/
--- a/.gitignore
+++ b/.gitignore
@@ -3,4 +3,8 @@
 .env.php
 .env
 .idea/
+
+# testing
 vendor/
+composer.lock
+.phpunit.cache/test-results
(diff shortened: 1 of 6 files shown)


recursivetree:
One difference to the old implementation is that we don't sort by relation. I don't quite understand why this has been done, and it also works without

PR #662, opened 2024-01-12
use bootstrap pagination templates

recursivetree:
During the upgrade to laravel 10, we missed that the css framework of the pagination views changed away from bootstrap 4. This PR tells laravel to use bootstrap 4.


recursivetree:
unit test failure seems to be unrelated

PR #665, opened 2024-01-25
change name of datatables js file

recursivetree:
Adapt to the breaking change in https://github.com/DataTables/Dist-DataTables/commit/5ef2c8a181736f29d83f25ddf88a0b8af157015b and let seat load the datatables js file againe


recursivetree:
please wait with merging, there are still a few broken things

PR #672, opened 2024-03-24
improve schedule seeder

recursivetree:
- This PR improves schedule seeding using the helper introduced here: https://github.com/eveseat/services/pull/179
- This other PR is required before this can be merged
- After the other PR is merged and released, we should bump the minimal services version of web


recursivetree:
CI failure seems to be unrelated, but we've been getting a lot of these failures recently. It might be worth investigating what is going wrong.

Crypta-Eve:
Yeah CI failure was due to the model factory not generating unique ids. Added to the list :P

PR #673, opened 2024-03-28
Fix null typeName

recursivetree:
Some of my corp mates fail to load the character sheet because a type is missing from the DB. In that case, just display `unknown`


setkeh:
I was able to manually make this change on my SEAT install and it corrected the issue with character sheets returning 500 internal errors.

PR #674, opened 2024-04-04
Squad Create form adds old type filters

moppa:
Only the Edit blade had been updated to use the new filter system. So if you added filters when you create your squad then the squad would not populate and related jobs that trigger the different squad observers failed.


recursivetree:
Something something I think I just forgot that when fixing the CVE originally :)
Anyways, I recommend that this should get merged.

The only thing I'm not sure about is whether it wouldn't be better to have one variable with the available filters instead of duplicating the same thing across multiple files. It might help with mistakes like this in the future.

PR #678, opened 2024-05-15
V5 link character use main char scopes

tehraven:
When a User clicks "Link Character" from the top-right menu, they should be prompted to use the same scopes as their Main Character.

## The Bug

The DB of refresh_tokens by user_id is not ordered naturally by created_at date, so the "first token" logic being replaced here pulls from a character other than Main Character, which means it will use potentially outdated defaults. Whereas the Main Character, being the one people actually use to Login with, is always prompted to use the newest default scopes that the SeAT Admin has chosen.

This fix allows SeAT Admins to change the default profile and provide an easy way for users with many characters linked to quickly updated all of the scopes across all characters.

## Before
- User has to log out of SeAT, log into SeAT with each and every character manually

## After
- User logs out and logs in with Main character, and then Link Character works fluidly thereafter using the new Scopes for the main character

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -36,7 +36,7 @@
     "guzzlehttp/guzzle": "^7.0",
     "intervention/image": "^2.0",
     "laravel/framework": "^10.0",
-    "laravel/horizon": "^5.0",
+    "laravel/horizon": "^5.24.4",
     "laravel/socialite": "^5.0",
     "laravel/ui": "^4.0",
     "lasserafn/php-initial-avatar-generator": "^4.2",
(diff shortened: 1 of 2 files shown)


Crypta-Eve:
Is there a reason for the bump in horizon version?

tehraven:
> Is there a reason for the bump in horizon version?

The original 5.0 release had supported PHP v7 and that led to some weird composer install outputs while developing a plugin that also tried to bring in module that used PHP 7.

Ref: https://github.com/laravel/horizon/commit/0b1bf46b21c397fdbe80b1ab54451fd227934508#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34L13

PR #682, opened 2024-07-04
Granular esi scheduling for users

recursivetree:
This PR builds upon https://github.com/eveseat/eveapi/pull/409 and allows admins to adjust the schedule of esi updates based on conditions.

Admins can specify an update interval associated with an acl role. From all roles a user holds, the one with the shortest update interval is used, and the characters associated with the user will be updated according to that update interval. If no scheduling rule is specified, or the user doesn't hold any roles, a default of one hour is used. This means after upgrading, nothing changes until it is configured to do so.

I've made the decision to bind the update interval to roles and therefore users, opposed to individual characters, because of multiple reasons:
- the filter system used by squads is computationally expensive. The performance impact of essentially duplicating the squads filter system is difficult to estimate. The solution implemented here avoids this by building on top of squad filters.
- the squads filter system is not really ready for use by something else without partially overhauling it.
- touching squads code must be done with care, as it is at the heart of the access control system. We cannot afford to get it wrong. This means more thorough code reviews etc and just more development time

Bildschirmfoto 2024-07-04 um 15 32 52


moppa:
@recursivetree I have the use case that I need to increase refresh rate for our main holding corporation toons for a number of users to get Notifications, Contracts etc for Alliance services.
[This ](https://github.com/eveseat/eveapi/pull/409) sounds exactly like what I need and I'd only need to change the `update_interval` in the DB.
However I see in this code that in certain scenarios here the `update_interval `gets updated based on the User roles. Would it be possible to add to the UI custom scheduling for certain toons and that those don't get overridden? I believe a lot of ppl have the same use case to get pings for shield timers and whatnot and need higher refresh on certain toons. Or would I just need to avoid touching this UI if I set custom `update_interval` from DB?

moppa:
As a reference my user has 40+ characters on SeAT and I need to get notifications more frequently on maybe 5 toons. Then there are a handful of other users with 10+ characters that has 1 character that need more frequent notifications and contracts etc.

recursivetree:
After some consideration, I've come to agree that is is worth it to have per-character instead of per-user scheduling. Therefore I'm going to close this and start a new PR. The eveapi side can remain unchanged.

PR #683, opened 2024-07-14
Logg out inactive users, and redirected them to the login page

justindh:
Currently a logged in but deactivated user can use seat and all of its plugins with their current permissions set. An inactive user that tries to login will receive an error. This change forces any user that is inactive, to be logged out and redirected to the login page.


tehraven:
I like this.

Seems to be some formatting issues, can you run the code formatter so you don't run into that issue with maintainers get around to reviewing this? (EX `if ($user->active == true){` without the space before bracket)

PR #702, opened 2025-02-11
fix one broken unit test

recursivetree:
As it stands right now, the CorporationInfo factory might generate `99000000` as alliance id with a 10% probability, but the test case assumes this never happens.

This is just one of the random testing failures. I'm sure there are more, but it is easier to track them down when they happen instead of proactively.


recursivetree:
funnily enough, the CI failure would be fixed by https://github.com/eveseat/eveapi/pull/429

PR #705, opened 2025-03-10
Refactor: trigger the event that leads to squad updates at RefreshToken level

recursivetree:
Changes:
- Introduces a `--sync` flag to `seat:filters:update` for debugging
- Rename the `CharacterFilterDataUpdate` event to `AuthedCharacterFilterDataUpdate `
- Trigger `AuthedCharacterFilterDataUpdate` with a `refresh_token` instead of a `character_info`. This allows all usages to get a user model even if the refresh token has already been deleted. Currently, all known usages covert the `character_info` back into a `refresh_token` or user via `refresh_token` anyways
- Only trigger `seat:filters:update` for characters who's token we own. For the current usages, it does not make sense to trigger them for other characters since 1) they can't be in a squad, 2) we don't run jobs for them
- These changes fix the race condition left open by the recent fix. `seat:filters:update` now removes user who's token is gone from squads, and `seat:filters:update` should run often enough to fix the issues cause by the race condition in an acceptable time given the low probability of encountering the race condition in the first place.

### DIFF
--- a/src/Commands/Seat/Filters/Update.php
+++ b/src/Commands/Seat/Filters/Update.php
@@ -35,7 +35,7 @@ class Update extends Command
      *
      * @var string
      */
-    protected $signature = 'seat:filters:update';
+    protected $signature = 'seat:filters:update {--sync}';
 
     /**
      * The console command description.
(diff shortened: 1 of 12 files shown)


recursivetree:
I still want to do some more testing before this can be merged.

